PR #3, opened 2023-06-07
Deploy Bedrock scripts

anikaraghu:
Add scripts for 
* Full L1 Bedrock Deploy process
* Deploy L1 bedrock implementation contracts
* Deploy L2 bedrock implementation contracts

### DIFF
--- a/script/deploy/l1/DeployBedrockL1ImplContracts.s.sol
+++ b/script/deploy/l1/DeployBedrockL1ImplContracts.s.sol
@@ -1,0 +1,118 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import "forge-std/Script.sol";
+import "forge-std/StdJson.sol";
+
+import "@eth-optimism-bedrock/contracts/L1/L1CrossDomainMessenger.sol";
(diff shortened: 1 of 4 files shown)


script/deploy/l1/DeployBedrockL1ImplContracts.s.sol
```diff
@@ -0,0 +1,118 @@
...
+    L1StandardBridge l1StandardBridgeImpl;
+    L2OutputOracle l2OutputOracleImpl;
+    OptimismPortal optimismPortalImpl;
+    OptimismMintableERC20Factory optimismMintableERC20FactoryImpl;
+    L1ERC721Bridge l1ERC721BridgeImpl;
+    SystemConfig systemConfigImpl;
+
+    function run() public {
+        utils = new Utils();
+        Utils.AddressesConfig memory addressCfg = utils.readAddressesFile();
```
anikaraghu, line 39: Worth nothing that with this approach, `addresses.json` will have to be added to the `inputs` folder for that deploy directory, which would mean `addresses.json` being duplicated in multiple places. Should be fine because none of the addresses should change, but an alternative would be that all addresses we use can be explicitly passed into the relevant script and defined in the `.env` file

PR #13, opened 2023-06-14
Challenger 1 of 2 Contract

anikaraghu:
Our Challenger role will be controlled by a 1 of 2 multisig with Base multisig and Optimism multisig as the signers. However if we use Gnosis safe for this, then any one signer could remove the other signer. We considered using some Gnosis safe plugins to enforce that signers cannot remove each other, but the logic for these was complex, and we decided it was safer and simpler to write a very simple contract to enforce this. 

## Testing
Solidity tests can be run with `forge test`.

Test txs: 
* Called `deleteL2Outputs()` through the test gnosis safe as a baseline test - hit [this error message](https://github.com/ethereum-optimism/optimism/blob/25b5f1ccf97b00cbfb7e214c12e79f5b50c677d7/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol#L144) as expected. [[tx link](https://goerli.etherscan.io/tx/0x4e129db1212625117139221c34a2856f38afb82bc1fa4f740042ba36f554cd66)]
image

* Deployed the MessagePasser smart contract [[tx link](https://goerli.etherscan.io/address/0x9Ea8E51Ed0Cf0E009969050eB47D269D287BDD2B)]

* Called `deleteL2Outputs()` through the smart contract via gnosis safe. Made it to L2OutputOracle contract and then reverted as expected. [[tx link](https://goerli.etherscan.io/tx/0xe5ef507f19454352bcfad46a2275736b6e1e7fd37ae8740899d3b9ea5d2e974f)]
image

* Called `deleteL2Outputs()` through the smart contract via a different gnosis which is not one of the whitelisted signers. Reverted at the MessagePasser level as expected. [[tx link](https://goerli.etherscan.io/tx/0x3901be5152dc64446268b79ac55faec990f19db62f604783f75aa1cc9972a34d)]
image

### DIFF
--- a/test/Challenger1of2.t.sol
+++ b/test/Challenger1of2.t.sol
@@ -4,8 +4,9 @@ pragma solidity ^0.8.15;
 import {console} from "forge-std/console.sol";
 import { Test, StdUtils } from "forge-std/Test.sol";
 
-import "@eth-optimism-bedrock/contracts/L1/L2OutputOracle.sol";
-import "@eth-optimism-bedrock/contracts/universal/ProxyAdmin.sol";
+import { L2OutputOracle } from "@eth-optimism-bedrock/contracts/L1/L2OutputOracle.sol";
+import { ProxyAdmin } from "@eth-optimism-bedrock/contracts/universal/ProxyAdmin.sol";
+import { Proxy } from "@eth-optimism-bedrock/contracts/universal/Proxy.sol";
 import { Address } from "@openzeppelin/contracts/utils/Address.sol";
 import { Challenger1of2 } from "src/Challenger1of2.sol";
 
@@ -33,6 +34,12 @@ contract Challenger1of2Test is Test {
         bytes _data,
         bytes _result
     );
+
+    event OutputsDeleted(
+        address indexed _caller,
+        uint256 indexed prevNextOutputIndex,
+        uint256 indexed newNextOutputIndex
+    );
     
     function setUp() public {
         vm.prank(deployer);
@@ -44,15 +51,15 @@ contract Challenger1of2Test is Test {
         );
 
         // Initialize L2OutputOracle implementation.
-        l2OutputOracle = new L2OutputOracle(
-            NONZERO_INTEGER,
-            NONZERO_INTEGER,
-            NONZERO_INTEGER,
-            ZERO,
-            proposer,
-            address(challenger),
-            NONZERO_INTEGER
-        );
+        l2OutputOracle = new L2OutputOracle({
+            _submissionInterval: NONZERO_INTEGER,
+            _l2BlockTime: NONZERO_INTEGER,
+            _startingBlockNumber: NONZERO_INTEGER,
+            _startingTimestamp: ZERO,
+            _proposer: proposer,
+            _challenger: address(challenger),
+            _finalizationPeriodSeconds: NONZERO_INTEGER
+        });
         
         vm.prank(deployer);
         // Upgrade and initialize L2OutputOracle.
@@ -115,16 +122,27 @@ contract Challenger1of2Test is Test {
     }
 
     function test_execute_opSigner_success() external {
-        proposeOutput();
-        proposeOutput();
+        _proposeOutput();
+        _proposeOutput();
 
         L2OutputOracle oracle = L2OutputOracle(address(l2OutputOracleProxy));
         // Check that the outputs were proposed.
         assertFalse(oracle.latestOutputIndex() == ZERO);
 
-        vm.prank(optimismWallet);
+        // We expect the OutputsDeleted event to be emitted
         vm.expectEmit(true, true, true, true, address(challenger));
+
+        // Emit the event we expect to see
         emit ChallengerCallExecuted(optimismWallet, DELETE_OUTPUTS_SIGNATURE, ZERO_OUTPUT);
+
+        // We expect deleteOutputs to be called
+        vm.expectCall(
+            address(l2OutputOracleProxy),
+            abi.encodeWithSignature("deleteL2Outputs(uint256)", 1)
+        );
+
+        // Make the call        
+        vm.prank(optimismWallet);
         challenger.execute(DELETE_OUTPUTS_SIGNATURE);
 
         // Check that the outputs were deleted.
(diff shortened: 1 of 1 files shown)


src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+     * Optimism or counter party signer).
+     * @param _data Data for function call.
+     */
+    function execute(bytes memory _data) external {
+        require(
+            msg.sender == OTHER_SIGNER || msg.sender == OP_SIGNER,
+            "Challenger1of2: must be an approved signer to execute"
+        );
+
+	    bytes memory result = Address.functionCall(
```
tynes: nit: extra whitespace

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+
+    /*//////////////////////////////////////////////////////////////
+                        External Functions
+    //////////////////////////////////////////////////////////////*/
+    /**
+     * @dev Executes a call as the Challenger (must be called by 
+     * Optimism or counter party signer).
+     * @param _data Data for function call.
+     */
+    function execute(bytes memory _data) external {
```
tynes, line 79: This could also be a fallback function that just passes thru the same calldata but whatever is easier

anikaraghu: Yep good callout, I considered it, but I like this because it feels a bit more explicit

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+        require(
+            msg.sender == OTHER_SIGNER || msg.sender == OP_SIGNER,
+            "Challenger1of2: must be an approved signer to execute"
+        );
+
+        bytes memory result = Address.functionCall(
+            L2_OUTPUT_ORACLE_PROXY,
+            _data,
+            "Challenger1of2: failed to execute"
+        );
```
tynes, line 89: Not sure what this does under the hood but assuming it just works and requires the call to succeed and passes all possible gas

test/Challenger1of2.t.sol
```diff
@@ -0,0 +1,163 @@
...
+        vm.prank(deployer);
+        proxyAdmin = new ProxyAdmin(deployer);
+        l2OutputOracleProxy = new Proxy(address(proxyAdmin));
+
+        challenger = new Challenger1of2(
+            optimismWallet, coinbaseWallet, address(l2OutputOracleProxy)
+        );
+
+        // Initialize L2OutputOracle implementation.
+        l2OutputOracle = new L2OutputOracle(
```
tynes: personal nit: prefer named args syntax when there is more than 1 arg to reduce cognitive overhead of reviewer

```
        new L2OutputOracle({
            _submissionInterval: submissionInterval,
            _l2BlockTime: 0,
            _startingBlockNumber: startingBlockNumber,
            _startingTimestamp: block.timestamp,
            _proposer: proposer,
            _challenger: owner,
            _finalizationPeriodSeconds: 7 days
        });
```

test/Challenger1of2.t.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        L2OutputOracle oracle = L2OutputOracle(address(l2OutputOracleProxy));
+        // Check that the outputs were proposed.
+        assertFalse(oracle.latestOutputIndex() == ZERO);
+
+        vm.prank(optimismWallet);
+        vm.expectEmit(true, true, true, true, address(challenger));
+        emit ChallengerCallExecuted(optimismWallet, DELETE_OUTPUTS_SIGNATURE, ZERO_OUTPUT);
+        challenger.execute(DELETE_OUTPUTS_SIGNATURE);
+
```
tynes, line 147: Could always do a `vm.expectCall` to ensure that the call is happening as well as a `vm.expectEmit` for the outputs deleted event
https://github.com/ethereum-optimism/optimism/blob/07b0dc2925fcc4019b7149ff60ccca8fcaff654a/packages/contracts-bedrock/contracts/L1/L2OutputOracle.sol#L166

We do see the side effect here tho

test/Challenger1of2.t.sol
```diff
@@ -0,0 +1,163 @@
...
+        vm.expectEmit(true, true, true, true, address(challenger));
+        emit ChallengerCallExecuted(coinbaseWallet, DELETE_OUTPUTS_SIGNATURE, ZERO_OUTPUT);
+        challenger.execute(DELETE_OUTPUTS_SIGNATURE);
+
+        // Check that the outputs were deleted.
+        assertEq(oracle.latestOutputIndex(), ZERO);
+    }
+
+    function proposeOutput() internal {
+        vm.startBroadcast(proposer);
```
tynes: personal nit: prefer `prank` over broadcast in tests to ensure that you dont accidentally send txs but its not that big of a deal

src/Challenger1of2.sol
anikaraghu: Yes, [this](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ecd2ca2cd7cac116f7a37d0e474bbb3d7d5e1c4d/contracts/utils/Address.sol#L128-L139) is the function for reference. If the function reverts, it'll bubble up the revert reason if there is one, or use the custom one provided here if there isn't (see [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ecd2ca2cd7cac116f7a37d0e474bbb3d7d5e1c4d/contracts/utils/Address.sol#L209)).

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+     * @dev The address of the L2OutputOracleProxy contract.
+     */
+    address public immutable L2_OUTPUT_ORACLE_PROXY;
+
+    /*//////////////////////////////////////////////////////////////
+                            EVENTS
+    //////////////////////////////////////////////////////////////*/
+    /**
+     * @dev Emitted when a Challenger call is made by a signer.
+     * @param _caller The signer making the call.
```
wilsoncusack, line 39: I have never under the the `_` prefix convention on args 🤔

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+                            Constructor
+    //////////////////////////////////////////////////////////////*/
+    /**
+     * @dev Constructor to set the values of the constants.
+     * @param _opSigner Address of Optimism signer.
+     * @param _otherSigner Address of counter party signer.
+     * @param _l2OutputOracleProxy Address of the L2OutputOracleProxy contract.
+     */
+    constructor(address _opSigner, address _otherSigner, address _l2OutputOracleProxy) {
+        require(_opSigner != address(0), "Challenger1of2: opSigner cannot be zero address");
```
wilsoncusack, line 59: out of scope for this, but seems we should align around using custom errors as it is more idiomatic to solidity these days, thoughts david?

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+        require(_otherSigner != address(0), "Challenger1of2: otherSigner cannot be zero address");
+        require(
+            _l2OutputOracleProxy.isContract(),
```
wilsoncusack, line 62: this was removed from latest https://github.com/OpenZeppelin/openzeppelin-contracts/commit/c5d040beb9a951b00e9cb57c4e7dd97cd04b45ac, can just be done with `.code.length > 0`

src/Challenger1of2.sol
wilsoncusack: do we have a test for the revert bubble up?

src/Challenger1of2.sol
```diff
@@ -0,0 +1,93 @@
...
+     */
+    address public immutable OP_SIGNER;
+
+    /**
+     * @dev The address of counter party's signer (likely a multisig)
+     */
+    address public immutable OTHER_SIGNER;
+
+    /**
+     * @dev The address of the L2OutputOracleProxy contract.
```
wilsoncusack, line 30: nit but could just be `///` if it's single line?

PR #19, opened 2023-06-15
check in FeeDisburser revenue share contract

dapperscene6:
The `FeeDisburser` contract will be the implementation contract behind a deterministically deployed `Proxy` on Base Mainnet which is identical to the following [`Proxy`](https://goerli.basescan.org/address/0x09C7bAD99688a55a2e83644BFAed09e62bDcCcBA#code) deployed on Base Goerli. Said `Proxy` contract  will be the `SequencerFeeVault`'s, `BaseFeeVault`'s, and `L1FeeVault`'s fee `RECIPIENT` on Base. It codifies the revenue share agreement between Base and Optimism and withdraws Base revenue to an `L1_WALLET`. 

## Testing
Solidity tests can be run with `forge test`
Code coverage can be checked with `forge coverage`

### DIFF
--- a/src/revenue-share/FeeDisburser.sol
+++ b/src/revenue-share/FeeDisburser.sol
@@ -104,7 +104,7 @@ contract FeeDisburser {
     ) {
         require(_optimismWallet != address(0), "FeeDisburser: OptimismWallet cannot be address(0)");
         require(_l1Wallet != address(0), "FeeDisburser: L1Wallet cannot be address(0)");
-        require(_feeDisbursementInterval > 0, "FeeDisburser: FeeDisbursementInterval cannot be zero");
+        require(_feeDisbursementInterval >= 24 hours, "FeeDisburser: FeeDisbursementInterval cannot be less than 24 hours");
 
         OPTIMISM_WALLET = _optimismWallet;
         L1_WALLET = _l1Wallet;
@@ -193,6 +193,10 @@ contract FeeDisburser {
         require(
             FeeVault(_feeVault).WITHDRAWAL_NETWORK() == FeeVault.WithdrawalNetwork.L2, 
             "FeeDisburser: FeeVault must withdraw to L2"
+        );
+        require(
+            FeeVault(_feeVault).RECIPIENT() == address(this), 
+            "FeeDisburser: FeeVault must withdraw to FeeDisburser contract"
         );
         if (_feeVault.balance >= FeeVault(_feeVault).MIN_WITHDRAWAL_AMOUNT()) {
             FeeVault(_feeVault).withdraw();
--- a/test/revenue-share/FeeDisburser.t.sol
+++ b/test/revenue-share/FeeDisburser.t.sol
@@ -40,7 +40,7 @@ contract FeeDisburserTest is CommonTest {
     // 101% denominated in basis points
     uint256 tooLargeBasisPoints = 10_001;
     // Denominated in seconds
-    uint256 feeDisbursementInterval = 60;
+    uint256 feeDisbursementInterval = 24 hours;
     uint256 minimumWithdrawalAmount = 10 ether;
     address proxyAdminOwner = address(2048);
 
@@ -78,11 +78,11 @@ contract FeeDisburserTest is CommonTest {
         new FeeDisburser(optimismWallet, payable(address(0)), feeDisbursementInterval);
     }
 
-    function test_constructor_fail_feeDisbursementInterval_Zero() external {
+    function test_constructor_fail_feeDisbursementInterval_lessThan24Hours() external {
         vm.expectRevert(
-            "FeeDisburser: FeeDisbursementInterval cannot be zero"
+            "FeeDisburser: FeeDisbursementInterval cannot be less than 24 hours"
         );
-        new FeeDisburser(optimismWallet, l1Wallet, ZERO_VALUE);
+        new FeeDisburser(optimismWallet, l1Wallet, 24 hours - 1);
     }
 
     function test_constructor_success() external {
@@ -109,8 +109,18 @@ contract FeeDisburserTest is CommonTest {
         feeDisburser.disburseFees();
     }
 
+    function test_disburseFees_fail_feeVaultWithdrawalToAnotherAddress() external {
+        sequencerFeeVault = new SequencerFeeVault(admin, minimumWithdrawalAmount, FeeVault.WithdrawalNetwork.L2);
+        vm.etch(Predeploys.SEQUENCER_FEE_WALLET, address(sequencerFeeVault).code);
+
+        vm.expectRevert(
+            "FeeDisburser: FeeVault must withdraw to FeeDisburser contract"
+        );
+        feeDisburser.disburseFees();
+    }
+
     function test_disburseFees_fail_minimumWithdrawalReversion() external {
-        FeeVaultRevert feeVaultRevert = new FeeVaultRevert();
+        FeeVaultRevert feeVaultRevert = new FeeVaultRevert(address(feeDisburser));
         vm.etch(Predeploys.SEQUENCER_FEE_WALLET, address(feeVaultRevert).code);
 
         vm.expectRevert(
--- a/test/revenue-share/mocks/FeeVaultRevert.sol
+++ b/test/revenue-share/mocks/FeeVaultRevert.sol
@@ -4,6 +4,16 @@ pragma solidity 0.8.15;
 import { FeeVault } from "@eth-optimism-bedrock/contracts/universal/FeeVault.sol";
 
 contract FeeVaultRevert {
+    address internal immutable _RECIPIENT;
+    
+    constructor(address _recipient) {
+        _RECIPIENT = _recipient;
+    }
+
+    function RECIPIENT() external view returns(address) {
+        return _RECIPIENT;
+    }
+    
     function WITHDRAWAL_NETWORK() external pure returns(FeeVault.WithdrawalNetwork) {
         return FeeVault.WithdrawalNetwork.L2;
     }


src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * @dev Emitted when fees are received from FeeVaults.
+     * @param _sender The FeeVault that sent the fees.
+     * @param _amount The amount of fees received.
+     */
+    event FeesReceived(address indexed _sender, uint256 _amount);
+    /**
+     * @dev Emitted when no fees are collected from FeeVaults at time of disbursement.
+     */
+    event NoFeesCollected();
```
tynes, line 89: Does this event need to be different than `FeesReceived` with a 0 `_amount`? Just thinking of making the abi smaller

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,206 @@
...
+     *          L1 Wallet Revenue Share = Gross Revenue - Optimism Revenue Share
+     */
+    function disburseFees() external virtual {
+        require(
+            block.timestamp >= lastDisbursementTime + FEE_DISBURSEMENT_INTERVAL,
+            "FeeDisburser: Disbursement interval not reached"
+        );
+        lastDisbursementTime = block.timestamp;
+
+        feeVaultWithdrawal(payable(Predeploys.SEQUENCER_FEE_WALLET));
```
tynes, line 133: Perhaps a comment here along the lines of calling the withdrawals here will mutate `netFeeRevenue` so that the control flow is more obvious. If you aren't familiar with the impl of the fee vault, it could make no sense as to why `netFeeRevenue` is used in arithmetic and then set to 0, it may look like its always 0

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,206 @@
...
+            return;
+        }
+
+        uint256 optimismGrossRevenueShare = feeBalance * OPTIMISM_GROSS_REVENUE_SHARE_BASIS_POINTS / BASIS_POINT_SCALE;
+
+        // Optimism's revenue share is the maximum of net and gross revenue
+        uint256 optimismRevenueShare = Math.max(optimismNetRevenueShare, optimismGrossRevenueShare);
+
+        // Send Optimism their revenue share on L2
+        Address.sendValue(OPTIMISM_WALLET, optimismRevenueShare);
```
tynes: fwiw we have the `SafeCall` library that we use for this kind of thing, to send value without risk of returndatabomb or passing too little gas, assuming the oz lib works well altho i have not looked at its impl

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,206 @@
...
+     */
+    function disburseFees() external virtual {
+        require(
+            block.timestamp >= lastDisbursementTime + FEE_DISBURSEMENT_INTERVAL,
+            "FeeDisburser: Disbursement interval not reached"
+        );
+        lastDisbursementTime = block.timestamp;
+
+        feeVaultWithdrawal(payable(Predeploys.SEQUENCER_FEE_WALLET));
+        feeVaultWithdrawal(payable(Predeploys.BASE_FEE_VAULT));
```
tynes, line 134: Could add requires to the fee vault withdrawals such that it reverts if the withdrawal is not going to the correct domain, this feels a bit safer to ensure that undefined states in the state machine cannot be reached

dapperscene6: @tynes do you mean adding a check to `feeVaultWithdrawal` that checks that the balance of `FeeDisburser` has increased the expected amount after calling `withdraw` on a `FeeVault`?

src/revenue-share/FeeDisburser.sol
dapperscene6: It definitely doesn't need to be different though I'm leaning towards keeping it as is as the ABI size increase seems worth the slight increase in clarity that comes from filtering for a completely separate event.

src/revenue-share/FeeDisburser.sol
dapperscene6: I'll switch this to use `SafeCall.send` as it's more configurable than `Address.sendValue`.

src/revenue-share/FeeDisburser.sol
tynes: What I mean is something like this:

```solidity
require(
    IFeeVault(Predeploys.SEQUENCER_FEE_WALLET).WITHDRAWAL_NETWORK() == WithdrawalNetwork.L2
);
```

Without this check, it is possible for the state machine to progress into an undefined state

dapperscene6: Makes sense, I'll add that check.

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,201 @@
...
+     */
+    constructor(
+        address payable _optimismWallet,
+        address _l1Wallet,
+        uint256 _feeDisbursementInterval
+    ) {
+        require(_optimismWallet != address(0), "FeeDisburser: OptimismWallet cannot be address(0)");
+        require(_l1Wallet != address(0), "FeeDisburser: L1Wallet cannot be address(0)");
+        require(_feeDisbursementInterval > 0, "FeeDisburser: FeeDisbursementInterval cannot be zero");
+
```
tynes, line 108: Can we require that the interval is exactly what it is in the contract? Or have it as a constant

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,201 @@
...
+    //////////////////////////////////////////////////////////////*/
+    /**
+     * @dev Withdraws funds from FeeVaults, sends Optimism their revenue share, and withdraws remaining funds to L1.
+     * @dev Implements revenue share business logic as follows:
+     *          Net Revenue             = sequencer FeeVault fee revenue + base FeeVault fee revenue
+     *          Gross Revenue           = Net Revenue + l1 FeeVault fee revenue
+     *          Optimism Revenue Share  = Maximum of Net Revenue and Gross Revenue
+     *          L1 Wallet Revenue Share = Gross Revenue - Optimism Revenue Share
+     */
+    function disburseFees() external virtual {
```
tynes, line 125: Why is this virtual?

src/revenue-share/FeeDisburser.sol
```diff
@@ -0,0 +1,201 @@
...
+    * @dev Withdraws fees from a FeeVault.
+    * @param _feeVault The address of the FeeVault to withdraw from.
+    * @dev Withdrawal will only occur if the given FeeVault's balance is greater than or equal to
+           the minimum withdrawal amount.
+    */
+    function feeVaultWithdrawal(address payable _feeVault) internal {
+        require(
+            FeeVault(_feeVault).WITHDRAWAL_NETWORK() == FeeVault.WithdrawalNetwork.L2, 
+            "FeeDisburser: FeeVault must withdraw to L2"
+        );
```
tynes, line 196: This would be more trustless if we also required the recipient was `address(this)`

test/revenue-share/FeeDisburser.t.sol
```diff
@@ -0,0 +1,294 @@
...
+        vm.etch(Predeploys.SEQUENCER_FEE_WALLET, address(feeVaultRevert).code);
+
+        vm.expectRevert(
+            "revert message"
+        );
+        feeDisburser.disburseFees();
+    }
+
+    function test_disburseFees_fail_withdrawalReversion() external {
+        vm.mockCall(Predeploys.SEQUENCER_FEE_WALLET, MINIMUM_WITHDRAWAL_AMOUNT_SIGNATURE, abi.encode(ZERO_VALUE));
```
maurelian, line 133: I'm not following why, if this call returns zero, we'd expect to see the error message below?

test/revenue-share/FeeDisburser.t.sol
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    function test_disburseFees_fail_minimumWithdrawalReversion() external {
+        FeeVaultRevert feeVaultRevert = new FeeVaultRevert();
+        vm.etch(Predeploys.SEQUENCER_FEE_WALLET, address(feeVaultRevert).code);
+
+        vm.expectRevert(
+            "revert message"
+        );
+        feeDisburser.disburseFees();
```
maurelian: This test seems to suggest that `disburseFees()` will revert if one of the `FeeVault` contracts has less than the minimum amount in it? But the implementation will actually just not call the contract in that situation, so I'm not clear about what this is testing for.

test/revenue-share/mocks/FeeVaultRevert.sol
```diff
@@ -0,0 +1,14 @@
...
+
+import { FeeVault } from "@eth-optimism-bedrock/contracts/universal/FeeVault.sol";
+
+contract FeeVaultRevert {
+    function WITHDRAWAL_NETWORK() external pure returns(FeeVault.WithdrawalNetwork) {
+        return FeeVault.WithdrawalNetwork.L2;
+    }
+
+    function MIN_WITHDRAWAL_AMOUNT() external pure returns(uint256) {
+        revert("revert message");
```
maurelian, line 22: This method should always return a value. It's the `withdraw()` function that can revert.

src/revenue-share/FeeDisburser.sol
dapperscene6: @tynes it's `virtual` incase we want to inherit from `FeeDisburser` and override it down the road with making code changes to the `FeeDisburser` contract.

src/revenue-share/FeeDisburser.sol
dapperscene6: @tynes happy to add this check as well 👍

test/revenue-share/FeeDisburser.t.sol
dapperscene6: @maurelian this tests how the `FeeDisburser` contract handles `MIN_WITHDRAWAL_AMOUNT` reverting. This is never expected to happen in practice.

test/revenue-share/FeeDisburser.t.sol
dapperscene6: @maurelian Yes, because our mock is returning a minimum withdrawal amount that's lower than the actual minimum withdrawal amount and so the call to `withdraw` fails.

test/revenue-share/mocks/FeeVaultRevert.sol
dapperscene6: @maurelian agreed. I still wanted to test what would happen if `MIN_WITHDRAWAL_AMOUNT` did revert. I'm also testing `withdraw()`'s reversion as well.

src/revenue-share/FeeDisburser.sol
dapperscene6: @tynes given that the `_feeDisbursementInterval` is the minimum amount of time that must occur between disbursements I'm happy to change the minimum value to 24 hours to match the contract. I'd like to keep it configurable though so as not to require code changes down the road in the event we want to change it. Does this work for you?

src/revenue-share/FeeDisburser.sol
tynes: Thank you!

src/revenue-share/FeeDisburser.sol
tynes: Got it, slight feeling of premature optimization because it sounds like there is a chance it will be inherited in the future and marking the function as virtual will not alter the bytecode and could be done at the time of the contract being inherited but its not that big of a deal either way

src/revenue-share/FeeDisburser.sol
tynes: Changing the minimum value works, we will need to verify the deployment anyways so we can just check for the value after its deployed

test/revenue-share/FeeDisburser.t.sol
maurelian: what is the desired behaviour of the `FeeDisburser` contract in that case? I don't really see any assertions about what it should do itself. It looks more like this is testing the behaviour of the mock contract?

test/revenue-share/FeeDisburser.t.sol
maurelian: I see, so the idea is to bypass this check to ensure that the call to the fee vault happens. 

https://github.com/base-org/contracts/blob/ae939fc304afffd3ca6de6fa177762c28eb7816c/src/revenue-share/FeeDisburser.sol#L201

This seems to me to be testing the fee vault and so is redundant with [this test](https://github.com/ethereum-optimism/optimism/blob/406e1257e61ad09edd5067f8ffa77a4931efec41/packages/contracts-bedrock/contracts/test/SequencerFeeVault.t.sol#L52-L58). Not a big issue, but also more to maintain.

test/revenue-share/FeeDisburser.t.sol
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    function test_constructor_fail_optimismWallet_ZeroAddress() external {
+        vm.expectRevert(
+            "FeeDisburser: OptimismWallet cannot be address(0)"
+        );
+        new FeeDisburser(payable(address(0)), l1Wallet, feeDisbursementInterval);
+    }
+
+    function test_constructor_fail_l1Wallet_ZeroAddress() external {
```
maurelian, line 74:
```suggestion
    function test_constructor_l1Wallet_ZeroAddress_reverts() external {
```

This would be more consistent with the [style guide for test names](https://github.com/ethereum-optimism/optimism/blob/406e1257e61ad09edd5067f8ffa77a4931efec41/packages/contracts-bedrock/STYLE_GUIDE.md#L117-L139).

test/revenue-share/FeeDisburser.t.sol
dapperscene6: It's testing that the `MIN_WITHDRAWAL_AMOUNT` error causes the entire `disburseFees()` transaction execution to fail.

test/revenue-share/FeeDisburser.t.sol
dapperscene6: That's the idea. The intention is not to test that the `FeeVault` reverts but to test how the `FeeDisburser` contract handles said reversion.

PR #20, opened 2023-06-15
Multisig + nested multisig builder library

mdehoog:
This PR introduces a couple of abstract contracts modeled from Optimism's [SafeBuilder](https://github.com/ethereum-optimism/optimism/blob/c7949cb3651c4ab07e44b26b1f2f95b795dc5e52/packages/contracts-bedrock/scripts/universal/SafeBuilder.sol#L4) library:

## `MultisigBuilder`
Much the same as `SafeBuilder`, but instead of submitting `approveHash` on chain, it generates an offchain signature for each signer, which should be sent to the final signer for transaction execution.

### Step 1: collect signatures from [threshold - 1] signers
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "sign()" | grep -A1 vvvvvvvv | grep -v vvvvvvvv | go run ./cmd/sign --private-key $SIGNER_1_PRIVATE_KEY
```
or with ledger:
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "sign()" | grep -A1 vvvvvvvv | grep -v vvvvvvvv | go run ./cmd/sign --ledger --hd-path "m/44'/60'/0'/0/0"
```

### Step 2: execute transaction (using signatures from step 1)
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "run(bytes)" 0x$SIGNATURES_CONCATENATED --private-key $SIGNER_2_PRIVATE_KEY --broadcast
```

## `NestedMultisigBuilder`
Similar to above, but assumes a nested multisig setup (executor multisig has owners who are themselves multisigs).

### Step 1: each approver safe ([nested safe threshold - 1]) must collect signatures from [threshold - 1] signers
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "signApproval(address)" $SAFE_1_ADDR | grep -A1 vvvvvvvv | grep -v vvvvvvvv | go run ./cmd/sign --private-key $SIGNER_1_PRIVATE_KEY
```

### Step2: final signer from each approver safe ([nested safe threshold - 1]) executes the transaction, using the signatures from step 1
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "runApproval(address,bytes)" $SAFE_1_ADDR 0x$SIGNATURES_FROM_STEP_1 --private-key $SIGNER_2_PRIVATE_KEY
```

### Step 3: the executor safe must collect signatures from [threshold - 1] signers
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "signTransaction(address)" $SAFE_2_ADDR | grep -A1 vvvvvvvv | grep -v vvvvvvvv | go run ./cmd/sign --private-key $SIGNER_3_PRIVATE_KEY
```

### Step 4: final signer from the executor safe executes the transaction, using the signatures from step 3
```bash
forge script --via-ir --rpc-url $RPC_URL MultisigBuilderTest --sig "runTransaction(address,bytes)" $SAFE_2_ADDR 0x$SIGNATURES_FROM_STEP_3 --private-key $SIGNER_4_PRIVATE_KEY
```

## Example usage
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.15;

import "./Test.sol";
import "../universal/MultisigBuilder.sol";

contract MultisigBuilderTest is MultisigBuilder { // or NestedMultisigBuilder
    function _postCheck() internal override view {
        // perform post execution checks
    }

    function _buildCalls() internal override view returns (IMulticall3.Call3[] memory) {
        IMulticall3.Call3[] memory calls = new IMulticall3.Call3[](1);

        uint256 value = 123;
        calls[0] = IMulticall3.Call3({
            target: 0x2e18754eA334c30Adb7Ecdd0A0d0D7dDf7A711E0,
            allowFailure: false,
            callData: abi.encodeCall(
                Test.setNumber,
                (value)
            )
        });

        return calls;
    }

    function _ownerSafe() internal override view returns (address) {
        return 0x5afad2BF5070995B61BB26223D0cDe780E596D95;
    }
}
```

### DIFF
--- a/script/universal/NestedMultisigBuilder.sol
+++ b/script/universal/NestedMultisigBuilder.sol
@@ -107,8 +107,8 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
         - <s7> => send signature to <s9>
         - <s8> => send signature to <s9>
      */
-    function signTransaction(address _safe, address[] memory _otherSafes) public returns (bool) {
-        IMulticall3.Call3 memory call = _generateExecuteCall(_safe, _otherSafes);
+    function signTransaction(address _safe) public returns (bool) {
+        IMulticall3.Call3 memory call = _generateExecuteCall(_safe);
         _printDataToSign(_safe, toArray(call));
         return true;
     }
@@ -128,9 +128,9 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
      * The following signer should run this function:
         - <s9> => using signatures from <s7>, <s8>
      */
-    function runTransaction(address _safe, address[] memory _otherSafes, bytes memory _signatures) public returns (bool) {
+    function runTransaction(address _safe, bytes memory _signatures) public returns (bool) {
         vm.startBroadcast();
-        IMulticall3.Call3 memory call = _generateExecuteCall(_safe, _otherSafes);
+        IMulticall3.Call3 memory call = _generateExecuteCall(_safe);
         bool success = _executeTransaction(_safe, toArray(call), _signatures);
         if (success) _postCheck();
         return success;
@@ -160,17 +160,31 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
         });
     }
 
-    function _generateExecuteCall(address _safe, address[] memory _otherSafes) internal returns (IMulticall3.Call3 memory) {
+    function _generateExecuteCall(address _safe) internal returns (IMulticall3.Call3 memory) {
         IGnosisSafe safe = IGnosisSafe(payable(_safe));
         address ownerSafe = _ownerSafe();
         IGnosisSafe nestedSafe = IGnosisSafe(payable(ownerSafe));
         bytes memory nestedData = _buildCalldata();
+        bytes32 nestedHash = _getTransactionHash(ownerSafe, nestedData);
 
-        address[] memory allSafes = new address[](_otherSafes.length + 1);
-        for (uint256 i; i < _otherSafes.length; i++) {
-            allSafes[i] = _otherSafes[i];
+        // get a list of owners that have approved this transaction
+        uint256 threshold = nestedSafe.getThreshold();
+        address[] memory owners = safe.getOwners();
+        address[] memory approvals = new address[](threshold);
+        uint256 approvalIndex;
+        for (uint256 i; i < owners.length; i++) {
+            address owner = owners[i];
+            uint256 approved = safe.approvedHashes(owner, nestedHash);
+            if (approved == 1) {
+                approvals[approvalIndex] = owner;
+                approvalIndex++;
+                if (approvalIndex == threshold - 1) {
+                    break;
+                }
+            }
         }
-        allSafes[_otherSafes.length] = _safe;
+        require(approvalIndex == threshold - 1, "not enough approvals");
+        approvals[approvalIndex] = _safe; // last approver is the current user
 
         return IMulticall3.Call3({
             target: ownerSafe,
@@ -186,7 +200,7 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
                     0,                           // gasPrice
                     address(0),                  // gasToken
                     payable(address(0)),         // refundReceiver
-                    addressSignatures(allSafes)  // signatures
+                    addressSignatures(approvals) // signatures
                 )
             )
         });


Makefile
```diff
@@ -10,7 +10,7 @@ install-foundry:
 
 .PHONY: solidity-deps
 solidity-deps: checkout-op-commit
```
anikaraghu, line 12: Do we need a command for building goland dependencies as well now?

cmd/sign/main.go
```diff
@@ -0,0 +1,187 @@
...
+		log.Fatalf("One (and only one) of --private-key, --ledger, --mnemonic must be set")
+	}
+
+	bytes, err := io.ReadAll(os.Stdin)
+	if err != nil {
+		log.Fatalf("Error reading from stdin: %v", err)
+	}
+	hash := common.FromHex(strings.TrimSpace(string(bytes)))
+	if len(hash) != 66 {
+		log.Fatalf("Expected EIP-712 hex string with 66 bytes, got %s", string(bytes))
```
anikaraghu: should we instead print the number of bytes, since not sure how useful the bytes string will be for people debugging?
```suggestion
		log.Fatalf("Expected EIP-712 hex string with 66 bytes, got %s bytes", len(hash))
```

Alternatively, we could first try to decode hex using `hex.DecodeString` (anyways [used under the surface here](https://github.com/ethereum/go-ethereum/blob/v1.12.0/common/bytes.go#L80)) and surface an error if there is one, and then check the length.

cmd/sign/main.go
```diff
@@ -0,0 +1,187 @@
...
+	if mnemonic != "" {
+		key, err := derivePrivateKey(mnemonic, path)
+		if err != nil {
+			return nil, fmt.Errorf("error deriving key from mnemonic: %w", err)
+		}
+		return &ecdsaSigner{key}, nil
+	}
+
+	// assume using a ledger
+	hub, err := usbwallet.NewLedgerHub()
```
anikaraghu: tiny nit, not sure `hub` is super clear to me
```suggestion
	ledgerHub, err := usbwallet.NewLedgerHub()
```

cmd/sign/main.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error starting ledger: %w", err)
+	}
+	wallets := hub.Wallets()
+	if len(wallets) == 0 {
+		return nil, fmt.Errorf("no legders found")
+	} else if len(wallets) > 1 {
+		return nil, fmt.Errorf("multiple ledgers found")
```
anikaraghu: Will this be the case if they have multiple funded accounts on their ledger? Can we provide the ability to specify which account to use? or is that already provided in the hdPath above?

cmd/sign/main.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	wallet := wallets[0]
+	if err := wallet.Open(""); err != nil {
+		return nil, fmt.Errorf("error opening ledger: %w", err)
```
anikaraghu: Will this error come up if the ledger is not unlocked? maybe if so, we say soemthign about unlocking their ledgger?

script/universal/NestedMultisigBuilder.sol
```diff
@@ -0,0 +1,186 @@
...
+        IMulticall3.Call3 memory call = _generateExecuteCall(_safe, _nestedSafe, _target, _otherSigners);
+        bool success = _executeTransaction(_safe, call, _signatures);
+        if (success) _postCheck(_target);
+        return success;
+    }
+
+    function _generateApproveCall(address _safe, address _nestedSafe, address _target) internal returns (IMulticall3.Call3 memory) {
+        IGnosisSafe safe = IGnosisSafe(payable(_safe));
+        IGnosisSafe nestedSafe = IGnosisSafe(payable(_nestedSafe));
+        bytes memory nestedData = buildCalldata(_target);
```
anikaraghu: nit: I think we call it just data in the `MultisigBuilder.sol` and it's the same process here. Calling it `nestedData` makes it feel like it has something to do with the nested safe which isn't really true, correct?
```suggestion
        bytes memory data = buildCalldata(_target);
```

script/universal/NestedMultisigBuilder.sol
```diff
@@ -0,0 +1,186 @@
...
+import { console } from "forge-std/console.sol";
+import { IMulticall3 } from "forge-std/interfaces/IMulticall3.sol";
+
+import { IGnosisSafe, Enum } from "@eth-optimism-bedrock/scripts/interfaces/IGnosisSafe.sol";
+import { EnhancedScript } from "@eth-optimism-bedrock/scripts/universal/EnhancedScript.sol";
+import { GlobalConstants } from "@eth-optimism-bedrock/scripts/universal/GlobalConstants.sol";
+
+/**
+ * @title NestedMultisigBuilder
+ * @notice Modeled from Optimism's SafeBuilder, but built for nested safes.
```
anikaraghu:
```suggestion
 * @notice Modeled from Optimism's SafeBuilder, but built for nested Safes (Safes where the signers are other Safes).
```

script/universal/NestedMultisigBuilder.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Step 1
+     * ======
+     * Generate a transaction approval data to sign. This method should be called by a threshold-1
+     * of members of each of the multisigs involved in the nested multisig, except the final
+     * multisig that will execute. Signers will pass their signature to the final signer of their
+     * repsective multisig.
+     *
+     * Example:
```
anikaraghu, line 53: thanks for the example and great documentation for each step!

script/universal/NestedMultisigBuilder.sol
```diff
@@ -0,0 +1,186 @@
...
+     * --------
+     * Given:
+     *  - 3-of-3 multisig <m1>, signers <s1>, <s2>, <s3>
+     *  - 2-of-3 multisig <m2>, signers <s4>, <s5>, <s6>
+     *  - 3-of-3 multisig <m3>, signers <s7>, <s8>, <s9>
+     *  - 3-of-3 nested multisig <n>, signers <m1>, <m2>, <m3>
```
anikaraghu: I wonder if this might make it a bit more clear?
```suggestion
     * A nested 3-of-3 multisig with the following <m1>, <m2>, <m3> multisigs as signers
     *  - 3-of-3 multisig <m1>, signers <s1>, <s2>, <s3>
     *  - 2-of-3 multisig <m2>, signers <s4>, <s5>, <s6>
     *  - 3-of-3 multisig <m3>, signers <s7>, <s8>, <s9>
```

script/universal/MultisigBuilder.sol
```diff
@@ -0,0 +1,74 @@
...
+        return true;
+    }
+
+    /**
+     * Step 2
+     * ======
+     * Execute the transaction. This method should be called by the final member of the
+     * multisig that will execute the transaction. Signatures from step 1 are required.
+     */
+    function run(address _safe, address _target, bytes memory _signatures) public returns (bool) {
```
anikaraghu: Thinking out loud - could we actually name this something besides `run()`, and then whoever writes the script the builds the calldata can implement their own `run`, which would call this function. The slightly awkward part of having the run implemented here is it becomes hard to customize what we pass in. For example, if we're making two calls and they have different targets, then we would have to hardcode just one of them in our script. It feels better to hardcode both, or better to not hardcode any and just pass the addrs in to a customized run function.

```suggestion
    function execute(address _safe, bytes memory _signatures) public returns (bool) {
```

cmd/sign/main.go
mdehoog: this is only if they have multiple ledgers connected at once, not multiple accounts. The accounts are selected via the hdPath

script/universal/NestedMultisigBuilder.sol
mdehoog: it is the nested data in this case... it's the data that the nested safe will receive

script/universal/MultisigBuilder.sol
mdehoog: i'll refactor to support multiple targets

Makefile
mdehoog: the idea of the golang script is to use it from `contract-deployments` rather than here, so i think just the source is needed in this repo

script/universal/MultisigBuilder.sol
mdehoog: changed to remove targets entirely, so it's up to the script implementation

script/universal/MultisigBase.sol
```diff
@@ -0,0 +1,128 @@
...
+    function addressSignatures(address[] memory _addresses) internal view returns (bytes memory) {
+        LibSort.sort(_addresses);
+        bytes memory signatures;
+        uint8 v = 1;
+        bytes32 s = bytes32(0);
+        for (uint256 i; i < _addresses.length; i++) {
+            bytes32 r = bytes32(uint256(uint160(_addresses[i])));
+            signatures = bytes.concat(signatures, abi.encodePacked(r, s, v));
+        }
+        return signatures;
```
anikaraghu, line 88: Im probably missing something, but I thought this format of signature only worked if the approvals were onchain?

mdehoog: yep correct, these are for the nested safe, which the approvals are onchain

PR #21, opened 2023-06-20
Add test contracts

anikaraghu:
Add a couple contracts which have been useful for testing in the past.


script/SafeBuilder.sol
```diff
@@ -1,203 +0,0 @@
-// SPDX-License-Identifier: MIT
```
anikaraghu, line 1: This was replaced by the contracts added in #20

script/calls/UnpausePortal.s.sol
```diff
@@ -1,38 +0,0 @@
-// SPDX-License-Identifier: MIT
```
anikaraghu, line 1: Cleanup - we agreed to keep scripts like this one in base-org/contract-deployments, since they are more specific to an operation.

PR #23, opened 2023-06-22
Nested Multisig Builder fix

anikaraghu:
I noticed when running step 3 in a test i.e. approvals in executor safe, I was getting `not enough approvals`. There were a couple places where I believe `safe` should have been `nestedSafe`. In this investigation, I noticed some namings that could have been a bit more clear, so made those changes as well

### DIFF
--- a/script/universal/NestedMultisigBuilder.sol
+++ b/script/universal/NestedMultisigBuilder.sol
@@ -64,9 +64,9 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
         - <s2> => send signature to <s3>
         - <s4> => send signature to <s5>
      */
-    function signApproval(address _safe) public returns (bool) {
-        IMulticall3.Call3 memory call = _generateApproveCall(_safe);
-        _printDataToSign(_safe, toArray(call));
+    function signApproval(address _signerSafe) public returns (bool) {
+        IMulticall3.Call3 memory call = _generateApproveCall();
+        _printDataToSign(_signerSafe, toArray(call));
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


script/universal/NestedMultisigBuilder.sol
```diff
@@ -153,40 +153,38 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
...
+    function _generateExecuteCall(address _signerSafe) internal returns (IMulticall3.Call3 memory) {
         address ownerSafe = _ownerSafe();
         IGnosisSafe nestedSafe = IGnosisSafe(payable(ownerSafe));
         bytes memory nestedData = _buildCalldata();
         bytes32 nestedHash = _getTransactionHash(ownerSafe, nestedData);
 
         // get a list of owners that have approved this transaction
         uint256 threshold = nestedSafe.getThreshold();
-        address[] memory owners = safe.getOwners();
+        address[] memory owners = nestedSafe.getOwners();
```
anikaraghu, line 182: this is one of the few actual code changes (not just rename)

script/universal/NestedMultisigBuilder.sol
```diff
@@ -153,40 +153,38 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
...
         address[] memory approvals = new address[](threshold);
         uint256 approvalIndex;
         for (uint256 i; i < owners.length; i++) {
             address owner = owners[i];
-            uint256 approved = safe.approvedHashes(owner, nestedHash);
+            uint256 approved = nestedSafe.approvedHashes(owner, nestedHash);
```
anikaraghu, line 187: this is the other actual code change (not just rename)

script/universal/NestedMultisigBuilder.sol
```diff
@@ -153,40 +153,38 @@ abstract contract NestedMultisigBuilder is EnhancedScript, GlobalConstants, Mult
         return abi.encodeCall(IMulticall3.aggregate3, (calls));
     }
 
-    function _generateApproveCall(address _safe) internal returns (IMulticall3.Call3 memory) {
-        IGnosisSafe safe = IGnosisSafe(payable(_safe));
-        address ownerSafe = _ownerSafe();
-        IGnosisSafe nestedSafe = IGnosisSafe(payable(ownerSafe));
+    function _generateApproveCall(address _signerSafe) internal returns (IMulticall3.Call3 memory) {
```
anikaraghu: this function doesn't actual used the passed in `_safe`

mdehoog: should we remove this parameter then? its been removed above

PR #24, opened 2023-06-26
Check-in the BalanceTracker contract

dapperscene6:
The BalanceTracker contract will be the implementation contract behind a [Proxy](https://github.com/ethereum-optimism/optimism/blob/a541c8a859d9258ad410598655f189de69adae19/packages/contracts-bedrock/contracts/universal/Proxy.sol) on Ethereum Mainnet. Said Proxy contract will receive fee withdrawals from L2 post revenue share with Optimism. The BalanceTracker will iterate through preconfigured system addresses, top the system addresses up to their target balances if they're below, and send remaining funds to a profit wallet.

### DIFF
--- a/src/revenue-share/BalanceTracker.sol
+++ b/src/revenue-share/BalanceTracker.sol
@@ -9,7 +9,7 @@ import { ReentrancyGuardUpgradeable }
 import { SafeCall } from "@eth-optimism-bedrock/contracts/libraries/SafeCall.sol";
 
 /**
-* @title BalanceTracker
+ * @title BalanceTracker
  * @dev Funds system addresses and sends the remaining profits to the profit wallet.
  */
 contract BalanceTracker is ReentrancyGuardUpgradeable {
@@ -48,7 +48,7 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
     /**
      * @dev Emitted when the BalanceTracker sends funds to a system address.
      * @param _systemAddress The system address being funded.
-     * @param _success A boolean denoting the success or failure of the fund send.
+     * @param _success A boolean denoting whether a fund send occurred and its success or failure.
      * @param _balanceNeeded The amount of funds the given system address needs to reach its target balance.
      * @param _balanceSent The amount of funds sent to the system address.
      */
@@ -158,14 +158,12 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
      */
     function refillBalanceIfNeeded(address _systemAddress, uint256 _targetBalance) internal {
         if (_systemAddress.balance >= _targetBalance) {
+            emit ProcessedFunds(_systemAddress, false, 0, 0);
             return;
         }
         
         uint256 valueNeeded = _targetBalance - _systemAddress.balance;
-        uint256 valueToSend = valueNeeded;
-        if (valueToSend > address(this).balance) {
-            valueToSend = address(this).balance;
-        }
+        uint256 valueToSend = valueNeeded > address(this).balance ? address(this).balance : valueNeeded;
 
         bool success = SafeCall.send(_systemAddress, gasleft(), valueToSend);
         emit ProcessedFunds(_systemAddress, success, valueNeeded, valueToSend);
--- a/test/revenue-share/BalanceTracker.t.sol
+++ b/test/revenue-share/BalanceTracker.t.sol
@@ -41,7 +41,6 @@ contract BalanceTrackerTest is CommonTest {
         vm.prank(proxyAdminOwner);
         balanceTrackerProxy.upgradeTo(address(balanceTrackerImplementation));
         balanceTracker = BalanceTracker(payable(address(balanceTrackerProxy)));
-        
     }
 
     function test_constructor_fail_profitWallet_zeroAddress() external {
@@ -217,6 +216,28 @@ contract BalanceTrackerTest is CommonTest {
         assertEq(l2OutputProposer.balance, ZERO_VALUE);
     }
 
+    function test_processFees_success_partialFunds() external {
+        uint256 partialBalanceTrackerBalance = INITIAL_BALANCE_TRACKER_BALANCE/3;
+        vm.deal(address(balanceTracker), partialBalanceTrackerBalance);
+        balanceTracker.initialize(
+            systemAddresses,
+            targetBalances
+        );
+        vm.expectEmit(true, true, true, true, address(balanceTracker));
+        emit ProcessedFunds(batchSender, true, batchSenderTargetBalance, partialBalanceTrackerBalance);
+        vm.expectEmit(true, true, true, true, address(balanceTracker));
+        emit ProcessedFunds(l2OutputProposer, true, l2OutputProposerTargetBalance, ZERO_VALUE);
+        vm.expectEmit(true, true, true, true, address(balanceTracker));
+        emit SentProfit(profitWallet, true, ZERO_VALUE);
+
+        balanceTracker.processFees();
+
+        assertEq(address(balanceTracker).balance, ZERO_VALUE);
+        assertEq(profitWallet.balance, ZERO_VALUE);
+        assertEq(batchSender.balance, partialBalanceTrackerBalance);
+        assertEq(l2OutputProposer.balance, ZERO_VALUE);
+    }
+
     function test_processFees_success_skipsAddressesAtTargetBalance() external {
         vm.deal(address(balanceTracker), INITIAL_BALANCE_TRACKER_BALANCE);
         vm.deal(batchSender, batchSenderTargetBalance);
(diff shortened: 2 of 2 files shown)


src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,191 @@
...
+     * @param _systemAddresses The system addresses being funded.
+     * @param _targetBalances The target balances for system addresses.
+     */
+    function initialize(
+        address payable[] memory _systemAddresses,
+        uint256[] memory _targetBalances
+    ) external initializer {
+        uint256 systemAddresesLength = _systemAddresses.length;
+        require(systemAddresesLength <= MAX_SYSTEM_ADDRESS_COUNT, 
+            "BalanceTracker: systemAddresses cannot have a length greater than 20");
```
anikaraghu, line 113: can we use `MAX_SYSTEM_ADDRESS_COUNT` in the error message instead of hardcoding?

src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,191 @@
...
+            return;
+        } 
+        
+        uint256 valueToSend = _targetBalance - _systemAddress.balance;
+        if (valueToSend > address(this).balance) {
+            valueToSend = address(this).balance;
+            emit NotEnoughFunds(_systemAddress, _systemAddress.balance + address(this).balance);
+        }
+
+        sendWithEvents(_systemAddress, valueToSend);
```
anikaraghu: tiny nit (might be more consistent with this function name):
```suggestion
        refillBalance(_systemAddress, valueToSend);
```

src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,191 @@
...
+     * @param _value The value being sent to the receiver address.
+     */
+    function sendWithEvents(address _receiver, uint256 _value) internal {
+        if (address(this).balance == 0) {
+            emit NoFundsToSend(_receiver);
+            return;
+        }
+
+        bool success = SafeCall.send(_receiver, gasleft(), _value);
+        if (success) {
```
wilsoncusack: Maybe a `processSend` event with a `success` flag?

src/revenue-share/BalanceTracker.sol
dapperscene6: To do this we'd either need to switch to a custom error or use string interpolation which is non-trivial in Solidity. Given that `MAX_SYSTEM_ADDRESS_COUNT` is a constant I'm leaning towards keeping it as is for now. That being said going forward I'm all for deviating from Optimism's use of `require` statements and defaulting to customer errors.

src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,173 @@
...
+                            Internal Functions
+    //////////////////////////////////////////////////////////////*/
+    /**
+     * @dev Checks the balance of the target address and refills it back up to the target balance if needed.
+     * @param _systemAddress The system address being funded.
+     * @param _targetBalance The target balance for the system address being funded.
+     */
+    function refillBalanceIfNeeded(address _systemAddress, uint256 _targetBalance) internal {
+        if (_systemAddress.balance >= _targetBalance) {
+            return;
```
wilsoncusack, line 162: I think you could just let this go through and you'd get an event with 0 needed and 0 sent?

dapperscene6: We'd still have to do a similar check to prevent an underflow on `L164`. I'd err towards keeping the short circuiting to prevent needless external contract calls. What do you think of emitting `ProcessedFunds` in the conditional so that we do emit `ProcessedFunds` for each system address every time `processFees` is called?

wilsoncusack: Hmm yeah good point. I think I'd prefer to keep on emit and wrap the send in this conditional, `_systemAddress.balance < _targetBalance`, but up to you. Seems worth an event here though, I think?

test/revenue-share/BalanceTracker.t.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        assertEq(balanceTracker.MAX_SYSTEM_ADDRESS_COUNT(), MAX_SYSTEM_ADDRESS_COUNT);
+        assertEq(balanceTracker.PROFIT_WALLET(), profitWallet);
+    }
+
+    function test_initializer_fail_systemAddresses_greaterThanMaxLength() external {
+        for (;systemAddresses.length <= balanceTracker.MAX_SYSTEM_ADDRESS_COUNT();) systemAddresses.push(payable(address(0)));
+        
+        vm.expectRevert(
+            "BalanceTracker: systemAddresses cannot have a length greater than 20"
```
wilsoncusack, line 69: any reason for this constraint?

test/revenue-share/BalanceTracker.t.sol
```diff
@@ -0,0 +1,272 @@
...
+        vm.expectRevert(
+            "BalanceTracker: systemAddresses cannot contain address(0)"
+        );
+        balanceTracker.initialize(
+            systemAddresses,
+            targetBalances
+        );
+    }
+
+    function test_initializer_fail_targetBalances_containsZero() external {
```
wilsoncusack, line 101: nit, mix of camel and snake case?

test/revenue-share/BalanceTracker.t.sol
```diff
@@ -0,0 +1,272 @@
...
+            targetBalances
+        );
+
+        assertEq(balanceTracker.systemAddresses(0), systemAddresses[0]);
+        assertEq(balanceTracker.systemAddresses(1), systemAddresses[1]);
+        assertEq(balanceTracker.targetBalances(0), targetBalances[0]);
+        assertEq(balanceTracker.targetBalances(1), targetBalances[1]);
+    }
+
+    function test_processFees_success_cannotBeReentered() external {
```
wilsoncusack, line 125: sorry where is reentracy getting blocked here?

test/revenue-share/BalanceTracker.t.sol
```diff
@@ -0,0 +1,272 @@
...
+
+        balanceTracker.processFees();
+
+        assertEq(address(balanceTracker).balance, ZERO_VALUE);
+        assertEq(profitWallet.balance, expectedProfitWalletBalance);
+        assertEq(batchSender.balance, batchSenderTargetBalance);
+        assertEq(l2OutputProposer.balance, l2OutputProposerTargetBalance);
+    }
+
+    function test_processFees_success_noFunds() external {
```
wilsoncusack, line 199: worth testing a best effort/partial send?

src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,173 @@
...
+        }
+        
+        uint256 valueNeeded = _targetBalance - _systemAddress.balance;
+        uint256 valueToSend = valueNeeded;
```
wilsoncusack: nbd but cleans up a lil and might save gas 
```suggestion
        uint256 valueToSend = valueNeeded > address(this).balance ? address(this).balance : valueNeeded;
```

wilsoncusack: not sure if worth writing address(this).balance to memory

src/revenue-share/BalanceTracker.sol
```diff
@@ -0,0 +1,173 @@
...
+        if (valueToSend > address(this).balance) {
+            valueToSend = address(this).balance;
+        }
+
+        bool success = SafeCall.send(_systemAddress, gasleft(), valueToSend);
```
wilsoncusack, line 168: this is nice for readability, could put this call in the emit to save gas writing to memory but 🤷

test/revenue-share/BalanceTracker.t.sol
wilsoncusack: Not sure I track reentrancy concern, also

dapperscene6: `processFees` has `nonReentrant` modifier. The failure mode is that if our first system address kept calling `processFees` it could drain all the `BalanceTracker`'s fund. Not a concern in our current setup but if we started funding addresses that we trust less it might be.

src/revenue-share/BalanceTracker.sol
dapperscene6: Going to keep it as is for readability

test/revenue-share/BalanceTracker.t.sol
dapperscene6: To prevent `BalanceTracker` from being initialized into a state where there are too many addresses for it to fund and so `processFees` will fail. Given we currently have only 2 addresses to fund 20 seemed like a safe upper bound.

test/revenue-share/BalanceTracker.t.sol
dapperscene6: I wrote the tests in a similar style to [Optimism's](I like the mix as IMO it increases test readability and started doing it to match Optimism's style). I'll add a TODO to completely match their tests style guide.

PR #26, opened 2023-07-18
Fix karalabe/usb MacOSX warning

mdehoog:
Fixed in https://github.com/karalabe/usb/pull/33 (not released yet)


cmd/sign/main.go
```diff
@@ -44,6 +44,7 @@ func main() {
 	}
 
 	bytes, err := io.ReadAll(os.Stdin)
+	fmt.Println()
```
mdehoog, line 47: improves the output slightly

PR #31, opened 2023-07-20
Add Tenderly simulation for signers using the NestedMultisigBuilder

mdehoog:
See https://dashboard.tenderly.co/shared/simulation/32e38115-ff94-469a-b875-1395af1192e6 as an example.

Signers should validate three things:
1. The `Nested hash:` value printed to the terminal matches the `hashToApprove` under `approveHash` and the `dataHash` when checking the signatures for the actual tx execution (in the above example, nested hash is `0x74ad46567c72c913b4922e50c2d3920dacf9da7038ba5f89024976352e9c270b`)
2. The `Data to sign:` value printed to the terminal (and showed on the ledger screen) matches the `data` field when checking the signatures for the `approveHash` tx (in the above example, data to sign is `0x19011b6ca1022dcb8ee88e2ab8dd1adf6f6bfa6af586f8b074e99c952e6e3c6befadd5c5183a56f8aa2520c9522029fc20aba61ce6c6aa32aec3b00519c95eb819e8`)
3. The actual state changes match what the script is intending to do (in the above case, it's a simple contract call to `0x2e18754ea334c30adb7ecdd0a0d0d7ddf7a711e0` for testing purposes)

Screenshots for above:
1. Screenshot 2023-07-20 at 8 27 43 AM
2. Screenshot 2023-07-20 at 8 28 04 AM
3. Screenshot 2023-07-20 at 8 28 20 AM

### DIFF
--- a/script/universal/Simulator.sol
+++ b/script/universal/Simulator.sol
@@ -74,46 +74,31 @@ abstract contract Simulator is CommonBase {
         );
         string memory username = abi.decode(userData, (string));
 
-        // url encoded JSON
+        // the following characters are url encoded: []{}
         string memory stateOverrides = "%5B";
         for (uint256 i; i < _overrides.length; i++) {
-            if (i > 0) {
-                stateOverrides = string.concat(
-                    stateOverrides,
-                    "%2C"
-                );
-            }
+            SimulationStateOverride memory _override = _overrides[i];
+            if (i > 0) stateOverrides = string.concat(stateOverrides, ",");
             stateOverrides = string.concat(
                 stateOverrides,
-                "%7B%22contractAddress%22%3A%22",
-                vm.toString(_overrides[i].contractAddress),
-                "%22%2C%22storage%22%3A%5B"
+                "%7B\"contractAddress\":\"",
+                vm.toString(_override.contractAddress),
+                "\",\"storage\":%5B"
             );
-            for (uint256 j; j < _overrides[i].overrides.length; j++) {
-                if (j > 0) {
-                    stateOverrides = string.concat(
-                        stateOverrides,
-                        "%2C"
-                    );
-                }
+            for (uint256 j; j < _override.overrides.length; j++) {
+                if (j > 0) stateOverrides = string.concat(stateOverrides, ",");
                 stateOverrides = string.concat(
                     stateOverrides,
-                    "%7B%22key%22%3A%22",
-                    vm.toString(_overrides[i].overrides[j].key),
-                    "%22%2C%22value%22%3A%22",
-                    vm.toString(_overrides[i].overrides[j].value),
-                    "%22%7D"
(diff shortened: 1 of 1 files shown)


script/universal/MultisigBase.sol
```diff
@@ -107,15 +106,19 @@ abstract contract MultisigBase is CommonBase, EnhancedScript {
...
-        bytes32 s = bytes32(0);
         for (uint256 i; i < _addresses.length; i++) {
-            bytes32 r = bytes32(uint256(uint160(_addresses[i])));
-            signatures = bytes.concat(signatures, abi.encodePacked(r, s, v));
+            signatures = bytes.concat(signatures, addressSignature(_addresses[i]));
         }
         return signatures;
     }
 
+    function addressSignature(address _address) internal pure returns (bytes memory) {
```
anikaraghu: Safe [refers to this](https://docs.safe.global/learn/safe-core/safe-core-protocol/signatures) as a "prevalidated signature", so maybe this?
```suggestion
    function prevalidatedSignature(address _address) internal pure returns (bytes memory) {
```

script/universal/NestedMultisigBuilder.sol
```diff
@@ -128,4 +129,80 @@ abstract contract NestedMultisigBuilder is MultisigBase {
...
+            )
+        );
+        calls[1] = IMulticall3.Call3({
+            target: _safe,
+            allowFailure: false,
+            callData: finalExec
+        });
+
+        SimulationStateOverride[] memory overrides = new SimulationStateOverride[](2);
+        // set the nested safe threshold to 1:
```
anikaraghu: This sounds kind of scary 😅 
```suggestion
        // The state change simulation sets the multisig threshold to 1 in the simulation to enable an approver to see what the final state change will look like upon transaction execution. The multisig threshold will not actually change in the transaction execution.
```

script/universal/Simulator.sol
```diff
@@ -0,0 +1,137 @@
...
+            for (uint256 j; j < _overrides[i].overrides.length; j++) {
+                if (j > 0) {
+                    stateOverrides = string.concat(
+                        stateOverrides,
+                        "%2C"
+                    );
+                }
+                stateOverrides = string.concat(
+                    stateOverrides,
+                    "%7B%22key%22%3A%22",
```
anikaraghu: are we able to use escape characters at least for some of these (like quotes) or does that cause issues? https://www.geeksforgeeks.org/solidity-strings/

script/universal/MultisigBase.sol
mdehoog: good call, done

script/universal/Simulator.sol
mdehoog: yep! replaced for all except `{}[]`

PR #34, opened 2023-07-29
Support submitting approveHash onchain, rather than signatures

mdehoog:
This will allow support for nested-nested-multisigs, e.g. there's a safe that is a signer for the nested safe (3 layers of safes). Instead of providing a signature, this inner safe can submit an `approveHash` to the middle safe, which will eventually allow `approveHash` on the outer safe.


script/universal/NestedMultisigBuilder.sol
```diff
@@ -70,6 +75,10 @@ abstract contract NestedMultisigBuilder is MultisigBase {
         address nestedSafeAddress = _ownerSafe();
         IMulticall3.Call3[] memory nestedCalls = _buildCalls();
         IMulticall3.Call3 memory call = _generateApproveCall(nestedSafeAddress, nestedCalls);
+
+        address[] memory approvers = _getApprovers(_signerSafe, toArray(call));
```
anikaraghu, line 79: Super helpful to automatically grab any onchain signatures! I was manually trying to construct the prevalidated signatures 😅 Could we add a comment mentioning that this happens automatically?

PR #39, opened 2023-08-14
Small documentation updates

anikaraghu:
* add SECURITY.md file
* update makefile commands to remove the word `solidity` since we don't do it for any other language


Sdsmetamask (commented):
👌

PR #44, opened 2023-08-24
Avoid printing long Tenderly URLs

mdehoog:
Tenderly's nginx has a URI limit of somewhere just over 8000 characters. This PR removes the `rawFunctionInput` query parameter if the URI is too long, and instead prints it separately for the user to copy/paste.

### DIFF
--- a/script/universal/Simulator.sol
+++ b/script/universal/Simulator.sol
@@ -109,14 +109,19 @@ abstract contract Simulator is CommonBase {
             vm.toString(block.chainid),
             "&contractAddress=",
             vm.toString(_to),
-            "&rawFunctionInput=",
-            vm.toString(_data),
             "&from=",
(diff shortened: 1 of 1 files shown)


script/universal/Simulator.sol
```diff
@@ -109,14 +109,19 @@ abstract contract Simulator is CommonBase {
...
-            "&rawFunctionInput=",
-            vm.toString(_data),
             "&from=",
             vm.toString(_from),
             "&stateOverrides=",
             stateOverrides
         );
-
-        console.log(str);
+        if (bytes(str).length + _data.length * 2 > 7980) {
```
mdehoog, line 117: `7980` == `8000 - "&rawFunctionInput=0x".length`

anikaraghu (approved):

PR #46, opened 2023-09-20
Smart Escrow Contract

anikaraghu:
Contract to handle transfer of OP tokens over vesting period


src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+    function initialTokens() public view virtual returns (uint256) {
+        return _initialTokens;
+    }
+
+    /**
+     * @dev Getter for number of tokens that vest after each vesting period
+     */
+    function vestingEventTokens() public view virtual returns (uint256) {
+        return _vestingEventTokens;
+    }
```
anikaraghu: We could alternatively choose to not have getters for these

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+        require(benefactorOwner != address(0), "SmartEscrow: benefactorOwner is zero address");
+        require(escrowOwner != address(0), "SmartEscrow: escrowOwner is zero address");
+        require(beneficiaryAddress != address(0), "SmartEscrow: beneficiaryAddress is zero address");
+        require(benefactorAddress != address(0), "SmartEscrow: benefactorAddress is zero address");
+
+        require(startTimestamp > 0, "SmartEscrow: start time is zero");
+        require(endTimestamp > startTimestamp, "SmartEscrow: end time is before start time");
+        require(vestingPeriodSeconds > 0, "SmartEscrow: vestingPeriodSeconds is zero");
+
+        // TODO: could add a check that checks the owners of the escrowOwner are beneficiaryOwner and benefactorOwner
```
anikaraghu: Thoughts on adding this check? Im a bit double minded about how general/specific to make this contract. In theory I could see other OP chains using this contract for token distribution as well, but maybe it's worth just making it specific to this use case

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+    event BenefactorUpdated(address oldBenefactor, address newBenefactor);
+    event ContractTerminated();
+
+    IERC20 public constant OP_TOKEN = IERC20(0x4200000000000000000000000000000000000042);
+    bytes32 public constant BENEFICIARY_OWNER = keccak256("smartescrow.roles.beneficiary");
+    bytes32 public constant BENEFACTOR_OWNER = keccak256("smartescrow.roles.benefactor");
+    bytes32 public constant ESCROW_OWNER = keccak256("smartescrow.roles.escrowowner");
+
+    address private _beneficiary;
+    address private _benefactor;
```
anikaraghu: open to other names here

src/smart-escrow/SmartEscrowOZ.sol
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @dev Allow beneficiary owner to update beneficiary address
+     *
+     * Emits a {BeneficiaryUpdated} event.
+     */
+    function updateBeneficiary(address newBeneficiary) external virtual onlyRole(BENEFICIARY_OWNER) {
+        require(newBeneficiary != address(0), "SmartEscrow: newBeneficiary is zero address");
+        if (beneficiary() != newBeneficiary) {
+            address oldBeneficiary = beneficiary();
+            _beneficiary = newBeneficiary;
```
anikaraghu: This actually doesn't work as is, since `_beneficiary` is private & immutable. We would likely need to use a new variable if we want to offer the ability to update it. Note: on OZ's master branch, they handle the beneficiary as an owner by using `Ownable` which means ownership could be changed, but this isn't yet included in a stable release and also using Ownable & AccessControl together could get confusing

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@openzeppelin/contracts/utils/Context.sol";
+
+/**
+ * @title SmartEscrow contract
+ * @dev Contract to handle payment of OP tokens over a period of vesting with
+ * the ability to terminate the contract.
+ */
+contract SmartEscrow is AccessControl {
+    event OPTransfered(uint256 amount, address recipient);
```
wilsoncusack: would index all the addresses here

src/smart-escrow/SmartEscrow.sol
wilsoncusack: I think better to just make specific to our use case

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+        withdrawUnvestedTokens();
+    }
+
+    /**
+     * @dev Allow beneficiary owner to update beneficiary address
+     *
+     * Emits a {BeneficiaryUpdated} event.
+     */
+    function updateBeneficiary(address newBeneficiary) external virtual onlyRole(BENEFICIARY_OWNER) {
+        require(newBeneficiary != address(0), "SmartEscrow: newBeneficiary is zero address");
```
wilsoncusack: nit on these errors, we had talked about migrating to custom errors as that's sort of latest convention.

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+        if (_beneficiary != newBeneficiary) {
+            address oldBeneficiary = _beneficiary;
+            _beneficiary = newBeneficiary;
+            emit BeneficiaryUpdated(oldBeneficiary, _beneficiary);
```
wilsoncusack: nit: avoid a storage load
```suggestion
            emit BeneficiaryUpdated(oldBeneficiary, newBeneficiary);
```

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+            address oldBenefactor = _benefactor;
+            _benefactor = newBenefactor;
+            emit BenefactorUpdated(oldBenefactor, _benefactor);
+        }
+    }
+
+    /**
+     * @dev Getter for the beneficiary address.
+     */
+    function beneficiary() public view virtual returns (address) {
```
wilsoncusack: why not just make it public, then, and not need the getter? 🤔

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,165 @@
...
+     * Emits a {BeneficiaryOwnerUpdated} event.
+     */
+    function updateBeneficiaryOwner(address newBeneficiaryOwner) external onlyOwner {
+        if (newBeneficiaryOwner == address(0)) revert AddressIsZeroAddress();
+        if (beneficiaryOwner != newBeneficiaryOwner) {
+            address oldBeneficiaryOwner = beneficiaryOwner;
+            beneficiaryOwner = newBeneficiaryOwner;
+            emit BeneficiaryUpdated(oldBeneficiaryOwner, newBeneficiaryOwner);
+        }
+    }
```
anikaraghu: I thought about making beneficiaryOwner immutable, but I do feel like there's cases in which we'd want to update it and having this done by the 2of2 owner seems right to me

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,165 @@
...
+    IERC20 public constant OP_TOKEN = IERC20(0x4200000000000000000000000000000000000042);
+
+    address public beneficiaryOwner;
+    address public beneficiary;
+    uint256 public released;
+    uint64 public immutable start;
+    uint64 public immutable end;
+    uint64 public immutable vestingPeriod;
+    uint64 private immutable _initialTokens;
+    uint64 private immutable  _vestingEventTokens;
```
anikaraghu: thoughts on private vs public for these?

src/smart-escrow/Errors.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+/// @dev The error is thrown when an address is not set.
+error AddressIsZeroAddress();
```
wilsoncusack: normally people just put these in the contract or interface itself

src/smart-escrow/SmartEscrow.sol
wilsoncusack: mm no strong opinion, also I think we can just use uint256 for everything for simplicity? It only saves you a little gas on deploy

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,165 @@
...
+        _transferOwnership(escrowOwner);
+    }
+
+    /**
+     * @dev Allow escrow owner (2-of-2 multisig with beneficiary and benefactor
+     * owners as signers) to terminate the contract.
+     *
+     * Emits a {ContractTerminated} event.
+     */
+    function terminate(address returnAddress) external onlyOwner {
```
wilsoncusack: these should be onlySuperAdmin or whatever? Or is do we control this?

anikaraghu: that's what `onlyOwner` should do?

src/smart-escrow/Errors.sol
anikaraghu: I've seen this pattern as well, but I can definitely just put it in the contract if preferred!

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * @title SmartEscrow contract
+ * @dev Contract to handle payment of OP tokens over a period of vesting with
+ * the ability to terminate the contract.
+ * This contract is inspired by OpenZeppelin's VestingWallet contract, but had sufficiently
+ * different requirements to where inheriting did not make sense.
+ */
+contract SmartEscrow is Ownable2Step {
+    event OPTransfered(uint256 amount, address indexed recipient);
```
wilsoncusack: would help to have some comments on these, or at least this one

src/smart-escrow/SmartEscrow.sol
wilsoncusack: so OP needs us to agree to terminate? Seems ok

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+        emit ContractTerminated();
+        withdrawUnvestedTokens(returnAddress);
+    }
+
+    /**
+     * @dev Allow contract owner to update beneficiary owner address.
+     *
+     * Emits a {BeneficiaryOwnerUpdated} event.
+     */
+    function updateBeneficiaryOwner(address newBeneficiaryOwner) external onlyOwner {
```
wilsoncusack: should we just let them update beneficiary, then? Do we really need beneficiaryOwner, if at the end of the day it's all controlled by owner?

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    /**
+     * @dev Release OP tokens that have already vested.
+     *
+     * Emits a {OPTransfered} event.
+     */
+    function release() public {
+        if (contractTerminated == true) revert ContractIsTerminated();
```
wilsoncusack: could drop == true

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+        uint256 amount = releasable();
+        if (amount > 0) {
+            released += amount;
+            emit OPTransfered(amount, beneficiary);
```
wilsoncusack: do we think we need this event 🤔

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/// @dev The error is thrown when an address is not set.
+error AddressIsZeroAddress();
+
+/// @dev The error is thrown when the start timestamp is zero.
+error StartTimeIsZero();
+
+/// @dev The error is thrown when the start timestamp is greater than the end timestamp.
+error StartTimeAfterEndTime(uint256 startTimestamp, uint256 endTimestamp);
```
anikaraghu, line 15: add Natspec to params

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+     * @dev Set initial parameters.
+     * @param beneficiaryOwnerAddress Address which can update the beneficiary address.
+     * @param beneficiaryAddress Address which receives tokens that have vested.
+     * @param escrowOwner Address which can terminate the contract.
+     * @param startTimestamp Timestamp of the start of vesting period (or the cliff, if there is one).
+     * @param endTimestamp Timestamp of the end of the vesting period.
+     * @param vestingPeriodSeconds Period of time between each vesting event in seconds.
+     * @param numInitialTokens Number of OP tokens which vest at start time.
+     * @param numVestingEventTokens Number of OP tokens which vest upon each vesting event.
+     */
```
refcell: Would be nice to have all comment styling consistent - preferably solidity natspec.

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,190 @@
...
+            emit OPTransfered(amount, returnAddress);
+            SafeERC20.safeTransfer(OP_TOKEN, returnAddress, amount);
+        }
+    }
+
+    /**
+     * @dev Getter for the amount of releasable OP.
+     */
+    function releasable() public view returns (uint256) {
+        return vestedAmount(block.timestamp) - released;
```
refcell, line 195: It's unfortunate we have to SSLOAD `released` twice. Instead, we might be able to change the implementation of `_vestingSchedule` so that it can be used directly by both `releasable` and `vestedAmount` separately.

refcell (commented):
Few comments

src/smart-escrow/SmartEscrow.sol
anikaraghu: this is an interesting insight! I'm trying to minimize the diff from the OZ Vesting wallet implementation, so am inclined to leave it as is, but let me know if you have a specific proposal

src/smart-escrow/SmartEscrow.sol
anikaraghu: It will be nice to have the flexibility to update the owner as needed, without relying on Optimism

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,182 @@
...
+
+    /// @notice Event emitted when the contract is terminated.
+    event ContractTerminated();
+
+    IERC20 public constant OP_TOKEN = IERC20(0x4200000000000000000000000000000000000042);
+
+    address public beneficiaryOwner;
+    address public beneficiary;
+    uint256 public released;
+    uint256 public immutable start;
```
wilsoncusack: nit, would put immutables below constant and above mutable

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,182 @@
...
+
+import "@openzeppelin/contracts/access/Ownable2Step.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/// @notice The error is thrown when an address is not set.
+error AddressIsZeroAddress();
+
+/// @notice The error is thrown when the start timestamp is zero.
+error StartTimeIsZero();
```
wilsoncusack: Errors usually defined in contract or interface

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,182 @@
...
+/// @title SmartEscrow contract
+/// @notice Contract to handle payment of OP tokens over a period of vesting with
+///         the ability to terminate the contract.
+/// @notice This contract is inspired by OpenZeppelin's VestingWallet contract, but had 
+///         sufficiently different requirements to where inheriting did not make sense.
+contract SmartEscrow is Ownable2Step {
+    /// @notice Event emitted when the beneficiary owner is updated.
+    /// @param oldOwner The address of the old beneficiary owner
+    /// @param newOwner The address of the new beneficiary owner
+    event BeneficiaryOwnerUpdated(address indexed oldOwner, address indexed newOwner);
```
wilsoncusack: official style guide is events go below state variables 
image
https://docs.soliditylang.org/en/latest/style-guide.html

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,182 @@
...
+    /// @param _beneficiaryOwner Address which can update the beneficiary address.
+    /// @param _beneficiary Address which receives tokens that have vested.
+    /// @param _escrowOwner Address which can terminate the contract.
+    /// @param _start Timestamp of the start of vesting period (or the cliff, if there is one).
+    /// @param _end Timestamp of the end of the vesting period.
+    /// @param _vestingPeriodSeconds Period of time between each vesting event in seconds.
+    /// @param _initialTokens Number of OP tokens which vest at start time.
+    /// @param _vestingEventTokens Number of OP tokens which vest upon each vesting event.
+    constructor(
+        address _beneficiaryOwner,
```
wilsoncusack: I know what you have here is common, but official solidity guide is to use a trailing when there is a collision. For parameters with no collision, I do not think we need any underscore  
image https://docs.soliditylang.org/en/latest/style-guide.html#avoiding-naming-collisions

anikaraghu: I was following OP's style guide here https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/STYLE_GUIDE.md#function-parameters

refcell: I appreciate the use of underscore-prefixed variables since it keeps things consistent and think this is the way to go.

src/smart-escrow/SmartEscrow.sol
```diff
@@ -0,0 +1,200 @@
...
+    /// @notice Allow contract owner to update beneficiary owner address.
+    /// @param _newBeneficiaryOwner Address to send remaining contract holdings to after termination
+    /// @notice Emits a {BeneficiaryOwnerUpdated} event.
+    function updateBeneficiaryOwner(address _newBeneficiaryOwner) external onlyOwner {
+        if (_newBeneficiaryOwner == address(0)) revert AddressIsZeroAddress();
+        if (beneficiaryOwner != _newBeneficiaryOwner) {
+            address oldBeneficiaryOwner = beneficiaryOwner;
+            beneficiaryOwner = _newBeneficiaryOwner;
+            emit BeneficiaryOwnerUpdated(oldBeneficiaryOwner, _newBeneficiaryOwner);
+        }
```
refcell: Since the happy path will more often than not be hit here (the new beneficiary owner is different than the old) we can save a `SLOAD` by moving the `oldBeneficiaryOwner` declaration above the `if` statement.

```suggestion
        address oldBeneficiaryOwner = beneficiaryOwner;
        if (oldBeneficiaryOwner != _newBeneficiaryOwner) {
            beneficiaryOwner = _newBeneficiaryOwner;
            emit BeneficiaryOwnerUpdated(oldBeneficiaryOwner, _newBeneficiaryOwner);
        }
````

File: src/smart-escrow/SmartEscrow.sol [hunk 25]
```diff
@@ -0,0 +1,200 @@
...
+    /// @param _newBeneficiary New beneficiary address
+    /// @notice Emits a {BeneficiaryUpdated} event.
+    function updateBeneficiary(address _newBeneficiary) external {
+        if (msg.sender != beneficiaryOwner) revert CallerIsNotOwner(msg.sender, beneficiaryOwner);
+        if (_newBeneficiary == address(0)) revert AddressIsZeroAddress();
+        if (beneficiary != _newBeneficiary) {
+            address oldBeneficiary = beneficiary;
+            beneficiary = _newBeneficiary;
+            emit BeneficiaryUpdated(oldBeneficiary, _newBeneficiary);
+        }
```
refcell: ibid

```suggestion
        address oldBeneficiary = beneficiary;
        if (oldBeneficiary != _newBeneficiary) {
            beneficiary = _newBeneficiary;
            emit BeneficiaryUpdated(oldBeneficiary, _newBeneficiary);
        }
```

src/smart-escrow/SmartEscrow.sol
refcell: Makes sense to me; it's fairly straightforward as is so may only be caught by auditing as a way to reduce gas if someone re-architects this

src/smart-escrow/SmartEscrow.sol
wilsoncusack: OK, thanks, as this isn't an OP Stack contract I'm not sure we need to use their style guide. Will ask about this particular style guide item in another forum.

anikaraghu:
Closing this - code pulled in to #51

PR #52, opened 2023-11-06
feat: override safe nonce and verify signatures

felipe-op:
This change adds a new environment variable called `SAFE_NONCE` which allow callers to override the Safe contract nonce.

This is useful when crafting future transactions ahead of time and collecting signatures.

It also provides a new `function verify(bytes memory _signatures)` to make sure the collected signatures are valid to execute the transaction.

It is intended to work as:
1. Maintainer craft a transaction ahead of time
2. Multiple signer parties are contacted and sign the transaction offline
3. Maintainer collect and verify the presigned transactions
4. Transaction with signatures are stored safely for later execution

### DIFF
--- a/script/universal/MultisigBase.sol
+++ b/script/universal/MultisigBase.sol
@@ -26,6 +26,13 @@ abstract contract MultisigBase is Simulator {
 
         IGnosisSafe safe = IGnosisSafe(payable(_safe));
         uint256 nonce = safe.nonce();
+        console.log("Safe current nonce:", nonce);
+
+        uint256 nonceOverride = vm.envUint("NONCE");
+        if (nonceOverride > nonce) {
+            nonce = nonceOverride;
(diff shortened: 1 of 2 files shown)


script/universal/MultisigBase.sol
```diff
@@ -26,6 +26,13 @@ abstract contract MultisigBase is Simulator {
 
         IGnosisSafe safe = IGnosisSafe(payable(_safe));
         uint256 nonce = safe.nonce();
+        console.log("Safe current nonce:", nonce);
+
+        uint256 nonceOverride = vm.envUint("SAFE_NONCE");
```
tynes, line 31:
```solidity
uint256 nonce = vm.envOr("SAFE_NONCE", safe.nonce());
```

mdehoog (approved):

PR #56, opened 2024-01-12
Optional SAFE_NONCE variable fix

anikaraghu:
The way we were using `envOr` to take in `SAFE_NONCE` as an optional argument was not compiling. After fixing, it was causing other issues across the codebase because we were using a `envOr` which was not marked as a view function in view functions. I checked with the foundry team and looks like this was an oversight so I upstreamed a change to make `envOr` a `view` function:
* https://github.com/foundry-rs/foundry/pull/6757
* https://github.com/foundry-rs/forge-std/pull/491

This now compiles & works (if you run into issues, try running `make deps` which should pull the latest forge-std version).

### DIFF
--- a/script/universal/MultisigBase.sol
+++ b/script/universal/MultisigBase.sol
@@ -28,8 +28,9 @@ abstract contract MultisigBase is Simulator {
         uint256 nonce = safe.nonce();
         console.log("Safe current nonce:", nonce);
 
-        if (bytes(vm.envOr("SAFE_NONCE", "")).length > 0) {
-            nonce = vm.envUint("SAFE_NONCE");
+        uint256 providedNonce = vm.envOr("SAFE_NONCE", nonce);
+        if (providedNonce > nonce) {
+            nonce = providedNonce;
             console.log("Creating transaction with nonce:", nonce);
         }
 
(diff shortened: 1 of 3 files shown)


mdehoog (dismissed):

script/universal/MultisigBase.sol
```diff
@@ -28,8 +28,9 @@ abstract contract MultisigBase is Simulator {
         uint256 nonce = safe.nonce();
         console.log("Safe current nonce:", nonce);
 
-        if (bytes(vm.envOr("SAFE_NONCE", "")).length > 0) {
-            nonce = vm.envUint("SAFE_NONCE");
+        uint256 providedNonce = vm.envOr("SAFE_NONCE", nonce);
```
mdehoog: alternative might just be `nonce = vm.envOr("SAFE_NONCE", nonce);` but this works

anikaraghu: yeah only thing was doesnt really make sense if the nonce is lower than the current nonce, so in that case we defualt to the current nonce here. but I could actually see someone maybe wanting to do that for testing (simulating some tx from the past)

OKEAMAH (approved):

cb-heimdall:
Review Error for OKEAMAH @ 2024-10-22 23:40:32 UTC
User must have write permissions to review

PR #57, opened 2024-01-23
fix: update contract imports for newer op commits; remove deploy scripts

cody-wang-cb:
The more recent op commits changed their code structure, so using the newer op commits cause import imcompatibility errors. Updating all the impacted imports in base contracts.

In addition, op also changed/removed some of their deploy related contracts so our deploy scripts don't work anymore. However they also added [foundry deploy scripts](https://github.com/ethereum-optimism/optimism/blob/e3ba24e72085d85bb5584dda33a03ccf60db86f0/packages/contracts-bedrock/scripts/Deploy.s.sol) which we can directly use in the future, thus deleting the deploy scripts in this repo. 

Ran `make test`, all tests passed
```
Ran 11 test suites: 82 tests passed, 0 failed, 0 skipped (82 total tests)
```


test/fee-vault-fixes/e2e/FeeVault.t.sol
```diff
@@ -2,10 +2,9 @@
 pragma solidity 0.8.15;
 
 import { CommonTest } from "test/CommonTest.t.sol";
-import { Predeploys } from "@eth-optimism-bedrock/contracts/libraries/Predeploys.sol";
-import { Proxy } from "@eth-optimism-bedrock/contracts/universal/Proxy.sol";
-import { FeeVault as FeeVault_Final } from "@eth-optimism-bedrock/contracts/universal/FeeVault.sol";
-import { L1FeeVault as L1FeeVault_Final } from "@eth-optimism-bedrock/contracts/L2/L1FeeVault.sol";
+import { Predeploys } from "@eth-optimism-bedrock/src/libraries/Predeploys.sol";
+import { Proxy } from "@eth-optimism-bedrock/src/universal/Proxy.sol";
+import { L1FeeVault as L1FeeVault_Final, FeeVault as FeeVault_Final } from "@eth-optimism-bedrock/src/L2/L1FeeVault.sol";
```
cody-wang-cb, line 7: Ran into some memory issues for `FeeVault.WithdrawalNetwork` when compiling (foundry 0.2.0) earlier
```
Error (7407): Type enum FeeVault.WithdrawalNetwork is not implicitly convertible to expected type enum FeeVault.WithdrawalNetwork.
  --> test/fee-vault-fixes/e2e/FeeVault.t.sol:27:29:
   |
27 |         withdrawalNetwork = L1FeeVault_Final(payable(Predeploys.SEQUENCER_FEE_WALLET)).WITHDRAWAL_NETWORK();
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This import change fixed the error now (reference: https://github.com/foundry-rs/foundry/issues/6407#issuecomment-1851226955)

script/universal/MultisigBase.sol
```diff
@@ -28,7 +28,7 @@ abstract contract MultisigBase is Simulator {
         uint256 nonce = safe.nonce();
         console.log("Safe current nonce:", nonce);
 
-        if (bytes(vm.envOr("SAFE_NONCE", "")).length > 0) {
+        if (bytes(vm.envOr(string("SAFE_NONCE"), string(""))).length > 0) {
```
cody-wang-cb, line 31: Not sure if this is the best fix, ran into this error earlier
```
Error (6675): Member "envOr" not unique after argument-dependent lookup in contract Vm.
  --> script/universal/MultisigBase.sol:31:19:
   |
31 |         if (bytes(vm.envOr("SAFE_NONCE", "")).length > 0) {
   |                   ^^^^^^^^

make: *** [test] Error 1
```

anikaraghu: Ah thanks this makes sense! I _think_ this would also work, but happy to just keep as is
```suggestion_
        if (vm.envOr(string("SAFE_NONCE"), 0)) > 0) {
```

anikaraghu: docs: https://book.getfoundry.sh/cheatcodes/env-or

anikaraghu (approved):
Thanks so much for all the fixes! 🎉

script/universal/MultisigBase.sol
cody-wang-cb: This gives the same error actually, pretty strange though. And no matter what I cast the 0 it doesn't work either. 
```
Error (6675): Member "envOr" not unique after argument-dependent lookup in contract Vm.
  --> script/universal/MultisigBase.sol:31:19:
   |
31 |         if (bytes(vm.envOr(string("SAFE_NONCE"), 0)).length > 0) {
```

So will keep it as is.

PR #60, opened 2024-02-09
feat: avoid unnecessary state overrides when nonces match

maurelian:
The current implementation will include an unnecessary nonce override if the NONCE env var is unset, or is set to the same as the current safe nonce value. 

2024-02-09 at 14 16@2x

This PR detects that condition, and excludes the nonce override in that case.

### DIFF
--- a/script/universal/Simulator.sol
+++ b/script/universal/Simulator.sol
@@ -28,12 +28,28 @@ abstract contract Simulator is CommonBase {
         });
     }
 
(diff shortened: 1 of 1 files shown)


script/universal/Simulator.sol
maurelian, line 1: I really tried to minimize the diff, but it's inherently a bit messy due to the need to allocate different sized arrays (if you allocate 2 storage overrides, there will be an empty override which will write to the 0th slot (overriding the implementation address). 

2024-02-09 at 14 24@2x

mdehoog (approved):

PR #65, opened 2024-03-08
Simulator override refactor

mdehoog:
Cleanup of simulator overrides to add them one-by-one. Also removes one unnecessary override (the `keccak256(0 || 2)` if `owner == address(0)`).


stevieraykatz:
Looks good -- works in an ad-hoc sim: https://www.tdly.co/shared/simulation/5df2eff3-42df-4de5-95d3-cdcde48e923c

stevieraykatz (approved):

PR #67, opened 2024-03-08
Add virtual method for explicit nonce overriding 

stevieraykatz:
Cleanup multiple instances of the same `try/catch` statement to look for .env nonce override. The new method is `virtual` so that inheriting scripts will have more flexibility over explicitly setting the Nonce value for a given safe transaction.


mdehoog (approved):
nice cleanup thanks ❤️

PR #72, opened 2024-04-10
feat: simulates transactions in forge and exposes state diff and sim info

mds1:
To facilitate the following features:
1. State diff assertions using state diffs from foundry.
2. Exposing transaction simulation payloads for simulating with external tools such as Tenderly.
3. Failing when a simulation or execution reverts or does not result in any state changes.

The below changes were made:
1. The postCheck hook takes state diff and simulation payload inputs.
2. The postCheck hook is mutable to enable ffi calls to further validate the transaction results.
3. Anytime a transaction is executed, whether simulated or broadcast, the state diff is recorded and returned to the caller.
4. The postCheck hook is always run after a transaction, whether simulated or broadcast.
5. Booleans returned from methods have been removed, as the methods will now revert if a transaction fails so they are no longer needed.

We had to copy out logic from the base repo in order to execute a simulation in foundry and make assertions on the state diff, as you can see here: https://github.com/ethereum-optimism/superchain-ops/pull/119/commits/815a5c42b99497a0de537aa28a7a658198926ba7

~~I'd still need to test this branch and make sure no changes were missed, and that it works properly to let us simplify that linked commit, but I wanted to open this draft PR to get feedback on the architecture from a high level and see if you'd be open to merging this PR.~~ — Edit: I've now tested this in the upstream https://github.com/ethereum-optimism/superchain-ops/pull/130 and it seems to be working!

### DIFF
--- a/script/universal/MultisigBuilder.sol
+++ b/script/universal/MultisigBuilder.sol
@@ -5,6 +5,7 @@ import "./MultisigBase.sol";
 
 import { console } from "forge-std/console.sol";
 import { IMulticall3 } from "forge-std/interfaces/IMulticall3.sol";
+import { Vm } from "forge-std/Vm.sol";
 
 /**
  * @title MultisigBuilder
@@ -20,7 +21,7 @@ abstract contract MultisigBuilder is MultisigBase {
     /**
      * @notice Follow up assertions to ensure that the script ran to completion.
      */
-    function _postCheck() internal virtual view;
+    function _postCheck(Vm.AccountAccess[] memory accesses, SimulationPayload memory simPayload) internal virtual;
 
     /**
      * @notice Creates the calldata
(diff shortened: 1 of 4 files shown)


script/universal/MultisigBuilder.sol
```diff
@@ -48,12 +49,12 @@ abstract contract MultisigBuilder is MultisigBase {
      * Alternatively, this method can be called by a threshold of signers, and those signatures
      * used by a separate tx executor address in step 2, which doesn't have to be a signer.
      */
-    function sign() public view returns (bool) {
```
maurelian, line 51: Is it OK to remove returns from all the public functions?

mds1: Touched on this in the PR description a bit but it seemed like they were never used, and all methods should now revert if a transaction fails, so my thinking was they don't add much value

stevieraykatz (approved):
Great work!

stevieraykatz (approved):

PR #76, opened 2024-04-17
Remove the `delayedVetoable` param from the `Vetoer1of2` constructor 

xenoliss:
This PR simplifies the deployment of the `Vetoer1of2` along with its associated `DelayedVetoable` contract by removing the circular dependency between both of them.

The `DELAYED_VETOABLE` immutable set at deployment in the `Vetoer1of2` is replaced by the additional `delayedVetoable`parameter that must be provided when calling the `execute` method.

Additionally, variables have been renamed to match with our [style guide](https://github.com/coinbase/solidity-style-guide).

### DIFF
--- a/src/Vetoer1of2.sol
+++ b/src/Vetoer1of2.sol
@@ -2,73 +2,80 @@
 pragma solidity 0.8.15;
 
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {DelayedVetoable} from "@eth-optimism-bedrock/src/L1/DelayedVetoable.sol";
 
-/**
- * @title Vetoer1of2
- * @dev This contract serves the role of the Vetoer, defined in DelayedVetoable.sol:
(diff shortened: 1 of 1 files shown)


nadir-akhtar-coinbase:
@xenoliss this looks good, though I'd prefer if we need not provide the `delayedVetoable` state variable every time we perform a veto call. A couple thoughts to make `delayedVetoable` permanent, either of which seems feasible to me:
1. Add a one-time initialize function that will 
2. Create the `delayedVetoable` contract here in the constructor of the `Vetoer1of2`, such that the `Vetoer1of2` will know the address of the `delayedVetoable` upon instantiation and vice versa

Moreover, generally regarding the contract (i.e. outside of this specific PR), I think we can continue making more changes that'll improve its use since we've made so many changes already (e.g. explicitly calling "veto()" rather than a generic "execute()"). Lmk if you're in agreement

nadir-akhtar-coinbase (approved):

PR #77, opened 2024-04-18
fix: snapshot and restore Safe nonce

mds1:
We recently merged https://github.com/base-org/contracts/pull/72. Prior to this PR, foundry did not simulate the transaction, and just logged the data needed to generate the tenderly sim. As a result, when forge called `safe.nonce()` at the very end of the solidity script, it matched what was on the actual chain.

In that PR, we modified the scripts so forge always applies the required state overrides and simulates the transaction. This was done so we can use the state diff recording cheats in foundry and make assertions on the diff.
However, the actual logging of the `Data to sign:` that's in the terminal calls `safe.nonce()` _after_ the foundry simulation, so it's one too high.

This PR snapshots the safe nonce prior to simulation and restores it after, so the data to sign is corrected logged.

### DIFF
--- a/script/universal/MultisigBase.sol
+++ b/script/universal/MultisigBase.sol
@@ -9,6 +9,7 @@ import "./Simulator.sol";
 
 abstract contract MultisigBase is Simulator {
     IMulticall3 internal constant multicall = IMulticall3(MULTICALL3_ADDRESS);
+    bytes32 internal constant SAFE_NONCE_SLOT = bytes32(uint256(5));
 
     function _getTransactionHash(address _safe, IMulticall3.Call3[] memory calls) internal view returns (bytes32) {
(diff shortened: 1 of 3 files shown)


mdehoog (approved):

script/universal/MultisigBase.sol
```diff
@@ -9,6 +9,7 @@ import "./Simulator.sol";
 
 abstract contract MultisigBase is Simulator {
     IMulticall3 internal constant multicall = IMulticall3(MULTICALL3_ADDRESS);
+    bytes32 internal constant SAFE_NONCE_SLOT = bytes32(uint256(5));
```
tynes, line 12: There is risk this will collide with another value that is automatically assigned by solc

PR #81, opened 2024-05-07
Add `cliffStart` to Smart Escrow contract 

stevieraykatz:
Per conversations with the OP team and CB legal, we have determined that the `initialTokens` "cliff amount" needs to be disbursed by the escrow contract at a later time than the start of the vesting schedule. As such, the logic to handle token release needs a new way to track the two different start times. 

This PR adds a `cliffStart` which strictly releases the `initialTokens`. The original `start` date will continue to track the release of tokens according to the vesting schedule. We strictly enforce that `start` <= `cliffStart` < `end`.


src/smart-escrow/SmartEscrow.sol
```diff
@@ -236,18 +253,30 @@ contract SmartEscrow is AccessControlDefaultAdminRules {
...
     /// @notice Returns the amount vested as a function of time.
     /// @param _timestamp The timestamp to at which to get the vested amount
     function _vestingSchedule(uint256 _timestamp) internal view returns (uint256) {
         if (_timestamp < start) {
             return 0;
-        } else if (_timestamp > end) {
-            return OP_TOKEN.balanceOf(address(this)) + released;
         } else {
-            return initialTokens + ((_timestamp - start) / vestingPeriod) * vestingEventTokens;
+            return ((_timestamp - start) / vestingPeriod) * vestingEventTokens;
```
anikaraghu: discussed offline - can make diff smaller by just changing original implementation to 
```
return _cliffRelease(_timestamp) + ((_timestamp - start) / vestingPeriod) * vestingEventTokens;
```

stevieraykatz: Implemented.

src/smart-escrow/SmartEscrow.sol
```diff
@@ -139,6 +153,8 @@ contract SmartEscrow is AccessControlDefaultAdminRules {
         }
         if (_start < block.timestamp) revert StartTimeCannotBeInPast(_start, block.timestamp);
         if (_start >= _end) revert StartTimeAfterEndTime(_start, _end);
+        if (_cliffStart < _start) revert CliffStartTimeInvalid(_cliffStart, _start);
+        if (_cliffStart >= _end) revert CliffStartTimeAfterEndTime(_cliffStart, _end);
```
wilsoncusack, line 157: nit: the naming of these errors kinda diverges in pattern. Should the first be called `CliffStartTimeBeforeStart`, though now that I type maybe it's awk

wilsoncusack (approved):
This seems good to me. Would be nice to try to think of some fuzz testing where we use `vm.warp`

cb-heimdall:
Review Error for wilsoncusack @ 2024-05-07 20:42:55 UTC
User failed mfa authentication, see [go/mfa-help](http://go/mfa-help)

src/smart-escrow/SmartEscrow.sol
```diff
@@ -242,12 +259,22 @@ contract SmartEscrow is AccessControlDefaultAdminRules {
...
+        }
+    }
+
+    /// @notice Returns the cliff amount if past the cliffStart time
+    /// @param _timestamp The timestamp to at which to get cliff release status 
+    function _cliffRelease(uint256 _timestamp) internal view returns (uint256) {
+        if (_timestamp < cliffStart) {
+            return 0;
+        } else {
+            return initialTokens;
```
nadir-akhtar-coinbase: Wondering why we need this function -- not sure why we can't simply return `initialTokens` above, since this should always return `initialTokens` based on how it's used here. The condition `_timestamp < cliffStart` is always expected to be true when this function is called from `_vestingSchedule`

stevieraykatz: Good catch! I implemented this before we added the strict time gating with start < cliffStart < end. Will remove and cleanup.

stevieraykatz:
> This seems good to me. Would be nice to try to think of some fuzz testing where we use `vm.warp`

I think that we've got some nice fuzzing implemented in `testFuzz_release`

nadir-akhtar-coinbase (approved):
As discussed with @stevieraykatz there's no way to validate upon construction that the total expected tokens (`initialTokens + vestingEventTokens * vestingEvents` where `vestingEvents = (end - start) / vestingPeriod`), meaning that there could be scenarios where `release()` reverts even though it's meant to distribute tokens. 

That said, 1) this is a topic separate from this PR, and 2) this is more of an operational concern than something that needs to be solved within the smart contract, as the worse case of a revert is no change

PR #84, opened 2024-05-31
feat: support nonce overrides for nested safes

mds1:
**Problem**: In the nested Safe case, we want to sign multiple transactions up-front, then execute them one after another. This requires overriding the safe nonce in advance. However, this is not currently supported for the nested Safe case.

**Solution**: In `NestedMultisigBuilder`, override the `_getNonce` method to change how we look for nonce overrides, and apply these overrides in `_simulateForSigner`. We cannot use the existing `SAFE_NONCE` env var since that is ambiguous in the nested context. Therefore the nested case now looks for env vars of the name `SAFE_NONCE_{UPPERCASE_SAFE_ADDRESS}` (example: `SAFE_NONCE_0X6DF4742A3C28790E63FE933F7D108FE9FCE51EA4`). While an uppercased address looks a little funny and isn't conventional, this preserves the UPPERCASE env var naming convention and sidesteps and issues around checksummed addresses. 

I am open to alternative solutions, but this felt like the simplest and most generic approach.

### DIFF
--- a/script/universal/NestedMultisigBuilder.sol
+++ b/script/universal/NestedMultisigBuilder.sol
@@ -40,6 +40,34 @@ abstract contract NestedMultisigBuilder is MultisigBase {
      * -----------------------------------------------------------
      */
 
+    // Virtual method which can be overwritten.
+    // This allows different nonce overrides for each safe in the nested multisig case.
+    // IMPORTANT: this method is used in the sign, simulate, AND execution contexts
+    // If you override it, ensure that the behavior is correct for all contexts
+    // As an example, if you are pre-signing a message that needs safe.nonce+1 (before safe.nonce is executed),
+    // you should explicitly set the nonce value with an env var.
+    // Overwriting this method with safe.nonce + 1 will cause issues upon execution because the transaction
+    // hash will differ from the one signed.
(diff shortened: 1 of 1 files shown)


script/universal/NestedMultisigBuilder.sol
```diff
@@ -40,6 +40,34 @@ abstract contract NestedMultisigBuilder is MultisigBase {
...
+        //   3. If it does not exist, we use the current nonce of the safe.
+        //   4. We explicitly do not use SAFE_NONCE as a fallback, because in the nested case it is
+        //      ambiguous which safe it refers to.
+        string memory safeNonceEnvVarName = string.concat("SAFE_NONCE_", vm.toUppercase(safeAddrStr));
+        try vm.envUint(safeNonceEnvVarName) {
+            nonce = vm.envUint(safeNonceEnvVarName);
+            console.log("Creating transaction with nonce:", nonce);
+        }
+        catch {}
+    }
```
stevieraykatz, line 69: With this schema it's only possible to create one overridden nonce per safe address. This strictly means that we can only pre-sign up to two transactions. Is this sufficient for what you need this for?

stevieraykatz (approved):

PR #123, opened 2025-01-30
(chore): Parameterize the BalanceTracker version

jackchuma:
This will allow new `BalanceTracker` versions to be initialized in the future.

Also updated an internal function name to include a leading underscore - in line with our style guide.

### DIFF
--- a/src/revenue-share/BalanceTracker.sol
+++ b/src/revenue-share/BalanceTracker.sol
@@ -13,13 +13,13 @@ import {SafeCall} from "@eth-optimism-bedrock/src/libraries/SafeCall.sol";
  */
 contract BalanceTracker is ReentrancyGuardUpgradeable {
     using Address for address;
+
     /*//////////////////////////////////////////////////////////////
                             Constants
     //////////////////////////////////////////////////////////////*/
     /**
      * @dev The maximum number of system addresses that can be funded.
      */
-
     uint256 public constant MAX_SYSTEM_ADDRESS_COUNT = 20;
 
     /*//////////////////////////////////////////////////////////////
@@ -37,6 +37,7 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
      * @dev The system addresses being funded.
      */
     address payable[] public systemAddresses;
+
     /**
      * @dev The target balances for system addresses.
      */
@@ -55,6 +56,7 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
     event ProcessedFunds(
         address indexed _systemAddress, bool indexed _success, uint256 _balanceNeeded, uint256 _balanceSent
     );
+
     /**
      * @dev Emitted when the BalanceTracker attempts to send funds to the profit wallet.
      * @param _profitWallet The address of the profit wallet.
@@ -62,6 +64,7 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
      * @param _balanceSent The amount of funds sent to the profit wallet.
      */
     event SentProfit(address indexed _profitWallet, bool indexed _success, uint256 _balanceSent);
+
     /**
      * @dev Emitted when funds are received.
      * @param _sender The address sending funds.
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

src/revenue-share/BalanceTracker.sol
```diff
@@ -91,10 +94,11 @@ contract BalanceTracker is ReentrancyGuardUpgradeable {
      * @dev Initializes the BalanceTracker contract.
      * @param _systemAddresses The system addresses being funded.
      * @param _targetBalances The target balances for system addresses.
+     * @param _version The version of the contract being initialized. Each new version must be greater than the previous version.
      */
-    function initialize(address payable[] memory _systemAddresses, uint256[] memory _targetBalances)
+    function initialize(address payable[] memory _systemAddresses, uint256[] memory _targetBalances, uint8 _version)
```
xenoliss, line 99: I can not see any access control on the `initialize()` function. Now that `_version` is a user-supplied parameter, this seems to allow arbitrary reinitialization by any caller who provides a higher version number.

test/revenue-share/BalanceTracker.t.sol
```diff
@@ -16,6 +16,7 @@ contract BalanceTrackerTest is CommonTest {
     event SentProfit(address indexed _profitWallet, bool indexed _success, uint256 _balanceSent);
     event ReceivedFunds(address indexed _sender, uint256 _amount);
 
+    uint8 constant VERSION = 1;
```
xenoliss, line 19: Would it be better to set it to 2 to match what was in the contract previously?

src/revenue-share/BalanceTracker.sol
cbfyi: tbh, perhaps we want a way to just change the addresses without having to call initialize

cbfyi: even if this was hardcoded (like before this PR), it'd have to be incremented every time we need to update the addresses, but nothing about the core implementation actually changed

xenoliss: My point is that the previous `reinitializer(2)` prevented it from being reinitialized. Now that it uses `reinitializer(_version)`, I think anyone could reinitialize the contract.

xenoliss: > tbh, perhaps we want a way to just change the addresses without having to call initialize

I think we would still need to enforce access control on who can update those addresses.

jackchuma: Wow yeah good catch. So you think we should make it ownable and add a setter function instead?

jackchuma: Probably not worth the trouble actually - inheriting a new contract could screw up the storage layout making it not feasible to upgrade. We'd have to add custom logic

PR #139, opened 2025-04-22
chore: use Call3Value for _buildCalls

xenoliss:
Change `_buildCalls` to return an array of `Call3Value`, enabling ETH transfers (primarily for funding new facilitator ledgers).

Update the unit tests with the new expected "data to sign", which changed due to the use of `aggregate3Value` instead of `aggregate3`, and the addition of the `value` field in the `Call3Value` calls.

PR #142, opened 2025-04-30
chore: add missing value to Call3Value calls

xenoliss:
This PR follows https://github.com/base/contracts/pull/139 and allow attaching a value to Multicall calls.

### DIFF
--- a/script/universal/MultisigScript.sol
+++ b/script/universal/MultisigScript.sol
@@ -226,8 +226,10 @@ abstract contract MultisigScript is Script {
             originalNonces[i] = _getNonce(_safes[i]);
         }
 
-        bytes[] memory datas = _transactionDatas(_safes);
-        (Vm.AccountAccess[] memory accesses, Simulation.Payload memory simPayload) = _simulateForSigner(_safes, datas);
+        (bytes[] memory datas, uint256 value) = _transactionDatas(_safes);
+
+        (Vm.AccountAccess[] memory accesses, Simulation.Payload memory simPayload) =
+            _simulateForSigner(_safes, datas, value);
 
         _postSign(accesses, simPayload);
         _postCheck(accesses, simPayload);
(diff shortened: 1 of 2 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

xenoliss:
I realized that in order to use the Multicall contract while transferring value from a Safe, the Multicall call cannot be a delegatecall. This is because a delegatecall cannot attach value to the message, and the `Multicall.aggregate3Value` function requires that `msg.value == sum(callValues)`.

A simple temporary fix is to allow performing non-delegated calls to the Multicall contract from our Safe multisig. The proposed implementation uses the following rule to determine when to perform a delegatecall vs a direct call:
 - If the call has no value attached to it, use a delegatecall.
 - If the call has a non-zero value, use a direct call.

This is a temporary solution, assuming the following invariant holds:
> We never need to upgrade from our owner Safe and transfer value in the same call.

If that becomes necessary, we will need to revise this logic, since a direct call to the Multicall changes `msg.sender`, which would break the upgrade mechanism.

mdehoog (approved):

PR #144, opened 2025-05-13
chore: improve multisig builder transaction simulation building

jackchuma:
Previously, our transaction simulation for MultisigBuilder setups required a workaround. To simulate a series of transactions through a hierarchy of multisigs using a single multicall, we had to temporarily change the ownership of the very first multisig in the chain to be solely the multicall contract. This allowed the multicall contract to initiate the first approval, which then cascaded up the chain, with each child multisig approving its parent, until the final multisig executed the actual task call(s). While this worked for most simulations, it prevented us from directly testing changes to the real signers of a multisig because our workaround had removed them. This update removes that ownership change. Instead, it directly grants an approval in the initial multisig on behalf of `msg.sender`, achieving the same simulation capability without losing the ability to manipulate actual signers.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

test/universal/DoubleNestedMultisigBuilder.t.sol
```diff
@@ -79,14 +79,20 @@ contract DoubleNestedMultisigBuilderTest is Test, DoubleNestedMultisigBuilder {
 
     function test_sign_double_nested_safe1() external {
         vm.recordLogs();
-        sign(safe1, safe3);
+        bytes memory txData = abi.encodeWithSelector(bytes4(keccak256("sign(address,address)")), safe1, safe3);
```
mdehoog: nit: could we use `abi.encodeCall` instead here and below? Or is there polymorphism blocking that?

script/universal/MultisigScript.sol
```diff
@@ -489,17 +494,17 @@ abstract contract MultisigScript is Script {
     // This allows simulation of the final transaction by overriding the threshold to 1.
     // State changes reflected in the simulation as a result of these overrides will
     // not be reflected in the prod execution.
-    function _safeOverrides(address _safe, address _owner)
+    function _safeOverrides(address _safe, address owner, bytes32 dataHash, uint256 index)
```
mdehoog: can we split this into 2 methods (one for index == 0 and one for not), and then do a ternary statement above? the conditional on the index is not super intuitive

or just get rid of this method entirely and call the `Simulation.` helpers directly

script/universal/Simulation.sol
```diff
@@ -54,29 +54,44 @@ library Simulation {
...
+        returns (StateOverride memory)
+    {
+        return addOverride(
+            _state,
+            StorageOverride({
+                key: keccak256(abi.encode(dataHash, keccak256(abi.encode(owner, uint256(8))))),
+                value: bytes32(uint256(0x1))
+            })
+        );
+    }
```
mdehoog, line 93: nice 😍

script/universal/MultisigScript.sol
jackchuma: Yup, actually `_safeOverrides` is basically doing that - so I just decided to inline this function where it was being called

test/universal/DoubleNestedMultisigBuilder.t.sol
```diff
@@ -79,14 +80,20 @@ contract DoubleNestedMultisigBuilderTest is Test, DoubleNestedMultisigBuilder {
 
     function test_sign_double_nested_safe1() external {
         vm.recordLogs();
-        sign(safe1, safe3);
+        bytes memory txData = abi.encodeCall(IDoubleNestedMultisigBuilder.sign, (safe1, safe3));
```
mdehoog: do you need the interface here (and below)? can you simply use the `DoubleNestedMultisigBuilder.sign` and remove `TestUtils.sol`?

jackchuma: Oh hmmmm, yeah it works if we change `sign` in the builder contracts to be `external` instead of `public` - do you have any opinions why we wouldn't want to change that?

script/universal/MultisigScript.sol
```diff
@@ -471,35 +473,35 @@ abstract contract MultisigScript is Script {
...
-            overrides[i] = _safeOverrides(_safes[i], owner);
+            uint256 nonce = _getNonce(_safes[i]);
+
+            if (i == 0) {
+                overrides[i] =
+                    Simulation.overrideSafeThresholdApprovalAndNonce(_safes[i], nonce, msg.sender, firstCallDataHash);
+            } else {
+                overrides[i] = Simulation.overrideSafeThresholdAndNonce(_safes[i], nonce);
+            }
         }
```
xenoliss: Could we do something like:
```suggestion
        overrides[0] = Simulation.overrideSafeThresholdApprovalAndNonce(_safes[0], _getNonce(_safes[0]), msg.sender, firstCallDataHash);

        for (uint256 i = 1; i < _safes.length; i++) {
            uint256 nonce = _getNonce(_safes[i]);
            overrides[i] = Simulation.overrideSafeThresholdAndNonce(_safes[i], nonce);
        }
```

script/universal/MultisigScript.sol
```diff
@@ -451,12 +453,12 @@ abstract contract MultisigScript is Script {
...
         private
-        pure
+        view
         returns (IMulticall3.Call3[] memory)
     {
         IMulticall3.Call3[] memory calls = new IMulticall3.Call3[](_safes.length);
         for (uint256 i; i < _safes.length; i++) {
-            address signer = i == 0 ? MULTICALL3_ADDRESS : _safes[i - 1];
+            address signer = i == 0 ? msg.sender : _safes[i - 1];
             // uint256 value = i == _safes.length - 1 ? _value : 0;
```
xenoliss: Not really relevant to this PR but might clean this as well (I left it unintentionally).

test/universal/DoubleNestedMultisigBuilder.t.sol
xenoliss: I can not think of any use case where we would need to do a internal call to sign in this script.

test/universal/DoubleNestedMultisigBuilder.t.sol
```diff
@@ -79,14 +79,20 @@ contract DoubleNestedMultisigBuilderTest is Test, DoubleNestedMultisigBuilder {
 
     function test_sign_double_nested_safe1() external {
         vm.recordLogs();
-        sign(safe1, safe3);
+        bytes memory txData = abi.encodeCall(DoubleNestedMultisigBuilder.sign, (safe1, safe3));
```
mdehoog, line 82: A potentially nicer syntax would be `DoubleNestedMultisigBuilder(address(this)).sign(safe1, safe3);`

mdehoog (approved):

PR #145, opened 2025-06-05
Feature: Add nested multisig deployment script

0xekkila:
- Add a script for deploying nested multisig based on a config json
- Add README for config json format
- Add related lib for the script


Makefile
```diff
@@ -43,3 +44,12 @@ bindings:
 	abigen --abi out/BalanceTracker.sol/BalanceTracker.abi.json --pkg bindings --type BalanceTracker --out bindings/balance_tracker.go
 	abigen --abi out/FeeDisburser.sol/FeeDisburser.abi.json --pkg bindings --type FeeDisburser --out bindings/fee_disburser.go
 	cd bindings && go mod tidy
+
+
+##
+# Script 
+## 
+.PHONY: deploy-nested
+deploy-nested:
```
jackchuma: I don't think we need this command in this repo - this script would likely be invoked from `contract-deployments`

remappings.txt
```diff
@@ -3,3 +3,4 @@
 @openzeppelin/contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/contracts
 @rari-capital/solmate/=lib/solmate/
 @solady/=lib/solady/src/
+@safe/=lib/safe-smart-account/contracts/
```
jackchuma: Let's remove this - it will increase complexity for our tasks in `contract-deployments`

script/universal/config/SAFE-CONFIG-README.md
jackchuma, line 1: Can we remove this `config` directory? It's not needed here. I think we should instead add a brief description at the top of `MultisigDeploy.sol` explaining what it does and include an example of the expected config json file format

jackchuma (commented):
Great job! Just a few things to clean up. Also, since we will mostly invoke this from a separate repo that inherits this repo as a dependency, I think it would be valuable to accept the expected file path to the config file as an env var instead of hardcoding it into the script itself

jackchuma (approved):

PR #153, opened 2025-07-28
fix: add array length validation in Recovery.withdrawETH

strmfos:
Add validation to ensure targets and amounts arrays have the same length in Recovery.withdrawETH method to prevent potential out-of-bounds access. This security improvement helps avoid panic errors that would occur if targets.length > amounts.length, and ensures correct function behavior by validating input parameters


jackchuma:
This added a test without actually updating the contract. It also is not needed. Thank you though!

PR #159, opened 2025-08-11
feat(CHAIN-1796): add CBMulticall contract

jackchuma:
Adds the Multicall contract utilized by our eth recovery task. In the regular `Multicall3` contract, the `aggregate3Value` function has the following line:

```solidity
require(msg.value == valAccumulator, "Multicall3: value mismatch");
```

This did not allow us to delegatecall from a multisig because you cannot send a value with this call type. Since our multisig already holds ETH, removing the above `require` statement allows calling `aggregate3Value` via delegatecall.

### DIFF
--- a/src/utils/CBMulticall.sol
+++ b/src/utils/CBMulticall.sol
@@ -1,0 +1,236 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+/// @title CBMulticall
+///
+/// @notice Aggregate results from multiple function calls
+///
+/// @dev This is an exact copy of the standard Multicall3 contract with one minor modification. The `require` statement
+///      at the bottom of `aggregate3Value` has been removed to support smart contract operations involving an array of
+///      permissioned calls that require a value set. When routing a call through a multisig, for example, it's
+///      beneficial to be able to DELEGATECALL from the multisig to this contract, thus maintaining the multisig as the
+///      `msg.sender` while still allowing the calls to utilize ETH already held by the multisig.
+contract CBMulticall {
+    struct Call {
+        address target;
+        bytes callData;
+    }
+
+    struct Call3 {
+        address target;
+        bool allowFailure;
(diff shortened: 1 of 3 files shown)


test/utils/CBMulticall.t.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    function test_blockAndAggregate_allSuccess_returnsResults() external {
+        CBMulticall.Call[] memory calls = new CBMulticall.Call[](2);
+        calls[0] =
+            CBMulticall.Call({target: address(target), callData: abi.encodeWithSelector(MockReceiver.bump.selector, 1)});
+        calls[1] =
+            CBMulticall.Call({target: address(target), callData: abi.encodeWithSelector(MockReceiver.bump.selector, 2)});
+        (uint256 bn, bytes32 bh, CBMulticall.Result[] memory res) = mc.blockAndAggregate(calls);
+        // silence unused vars while still asserting one result
```
xenoliss: Can be removed

src/utils/CBMulticall.sol
```diff
@@ -0,0 +1,236 @@
...
+                }
+            }
+            unchecked {
+                ++i;
+            }
+        }
+    }
+
+    /// @notice Aggregate calls with a msg value
+    /// @notice Reverts if msg.value is less than the sum of the call values
```
xenoliss: I think we should also remove this comment now.

src/utils/CBMulticall.sol
```diff
@@ -0,0 +1,236 @@
...
+    /// @param calls An array of Call3Value structs
+    /// @return returnData An array of Result structs
+    function aggregate3Value(Call3Value[] calldata calls) public payable returns (Result[] memory returnData) {
+        uint256 valAccumulator;
```
xenoliss: Not used anymore

xenoliss (approved):

PR #161, opened 2025-10-12
feat: optional state diff output

jackchuma:
Adds a `StateDiff` library that facilitates optional state diff recording to a local file after transaction simulation. This allows us to analyze the transaction state diff using forge cheatcodes instead of having to maintain our own custom evm database implementation. Tasks that wish to use this feature need to set `RECORD_STATE_DIFF=true` in their `.env` file

### DIFF
--- a/script/universal/MultisigScript.sol
+++ b/script/universal/MultisigScript.sol
@@ -237,7 +237,7 @@ abstract contract MultisigScript is Script {
         bytes memory txData = _encodeTransactionData({safe: safes[0], data: datas[0], value: value});
         StateDiff.recordStateDiff({json: json, parents: parents, txData: txData, targetSafe: _ownerSafe()});
 
-        _printDataToSign({safe: safes[0], data: datas[0], value: value});
+        _printDataToSign({safe: safes[0], data: datas[0], value: value, txData: txData});
     }
 
     /// Step 1.1 (optional)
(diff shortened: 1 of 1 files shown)


script/universal/StateDiff.sol
```diff
@@ -0,0 +1,86 @@
...
+        bytes memory txData,
+        address targetSafe
+    ) internal {
+        json = VM.serializeBytes(OBJ, "preimages", abi.encode(parents));
+        json = VM.serializeBytes(OBJ, "dataToSign", txData);
+        json = VM.serializeAddress(OBJ, "targetSafe", targetSafe);
+
+        bool shouldWrite = VM.envOr("RECORD_STATE_DIFF", false);
+        if (shouldWrite) {
+            VM.writeJson(json, "stateDiff.json");
```
leopoldjoy, line 70: Would it make sense to include some unique information in the file name so that it doesn't overwrite existing output files? (E.g. block number, tx hash, etc.)

script/universal/MultisigScript.sol
```diff
@@ -228,6 +233,10 @@ abstract contract MultisigScript is Script {
         for (uint256 i; i < safes.length; i++) {
             vm.store({target: safes[i], slot: SAFE_NONCE_SLOT, value: bytes32(originalNonces[i])});
         }
+
+        bytes memory txData = _encodeTransactionData({safe: safes[0], data: datas[0], value: value});
```
leopoldjoy, line 237: This function, `_encodeTransactionData()` is called again inside the `_printDataToSign` function call 3 lines down ([here](https://github.com/base/contracts/pull/161/files#diff-72074ac05b528d57b61e77e3e0c6796a09cd56c4966a3d88a2c086b4a539ffceR240)). It might be more efficient to pass in `txData` to the `_printDataToSign` function call.

script/universal/StateDiff.sol
jackchuma: Hmm, so my thinking is it should produce a deterministic file name so the validation tool knows where to look, and the tool will delete the file after processing. One idea is we could use block number or timestamp, but also do a `stateDiff-latest.json` similar to foundry execution artifacts, but that seems like overkill to me. I'm open to it if you're opinionated that we should not delete the files after processing though

PR #167, opened 2025-11-11
feat(CHAIN-2551): modify MultisigScript for OPCM tasks

jackchuma:
Modifies `MultisigScript` to support DELEGATECALL operations to OPCM

### DIFF
--- a/test/universal/MultisigScriptDelegateCall.t.sol
+++ b/test/universal/MultisigScriptDelegateCall.t.sol
@@ -108,4 +108,3 @@ contract MultisigScriptDelegateCallTest is Test, MultisigScript {
     }
 }
 
-


script/universal/MultisigScript.sol
```diff
@@ -147,6 +147,13 @@ import {StateDiff} from "./StateDiff.sol";
 ///     │        │        │        │        │        │          │            run()             │
 ///     │        │        │        │        │        │          │─────────────────────────────>│
 abstract contract MultisigScript is Script {
+    struct SafeTx {
```
jackchuma: Needed to add this to encapsulate inputs for one of the internal functions to avoid a stack too deep error

script/universal/MultisigScript.sol
```diff
@@ -267,9 +285,10 @@ abstract contract MultisigScript is Script {
     /// @param signatures The signatures from step 1 (concatenated, 65-bytes per sig)
     function approve(address[] memory safes, bytes memory signatures) public {
         safes = _appendOwnerSafe({safes: safes});
-        (bytes[] memory datas, uint256 value) = _transactionDatas({safes: safes});
-        (Vm.AccountAccess[] memory accesses, Simulation.Payload memory simPayload) =
-            _executeTransaction({safe: safes[0], data: datas[0], value: value, signatures: signatures, broadcast: true});
+        (bytes[] memory datas, uint256 value,) = _transactionDatas({safes: safes});
+        (Vm.AccountAccess[] memory accesses, Simulation.Payload memory simPayload) = _executeTransaction({
+            safe: safes[0], to: multicallAddress, data: datas[0], value: value, signatures: signatures, broadcast: true
```
jjtny1: Should this `to:` still be `multicallAddress`?

jackchuma: I believe so, yes. Only the final execute call should be directed to OPCM. The approval flow still uses multicall

jjtny1 (approved):

PR #170, opened 2025-11-24
Refactor MultisigScript

xenoliss:
This PR refactors the `MultisigScript` logic to remove unnecessary `Multicall` wrapping, provide more flexibility when building script calls, and use our custom `CBMulticall` instead of the canonical implementation.

The `_buildCalls()` method now returns an array of `Call` structs:

```solidity
struct Call {
    Enum.Operation operation;
    address target;
    bytes data;
    uint256 value;
}
```

Each call specifies its own execution mode (`Call` or `DelegateCall`) via the `operation` field. The internal `_buildCallsChecked()` method validates these calls, preventing invalid combinations such as non-zero `value` for delegate calls.

The Safe-to-Safe approval chain has been simplified: approvals are now direct `approveHash` calls without any multicall wrapping.

The only remaining multicall wrapping occurs when `_buildCalls()` returns multiple calls. If there is exactly one call, it is executed directly. When multiple calls exist, `_buildAggregatedScriptCall()` groups and aggregates compatible calls. For example, the following calls returned by `_buildCalls()`:

```solidity
[
    Call(Call, ContractA, abi.encodeCall(a, ()), 0),
    Call(Call, ContractB, abi.encodeCall(b, ()), 0),
    Call(DelegateCall, OPCM, abi.encodeCall(upgrade, ()), 0),
    Call(Call, ContractC, abi.encodeCall(c, ()), 100),
    Call(Call, ContractD, abi.encodeCall(d, ()), 200)
]
```

are aggregated into:

```solidity
CBMulticall.aggregateDelegateCalls(
    Call3(
        target = Multicall,
        allowFailure = false,
        calldata = abi.encodeCall(Multicall.aggregate3, (
            Call3(ContractA, false, abi.encodeCall(a, ())),
            Call3(ContractB, false, abi.encodeCall(b, ()))
        ))
    ),
    Call3(
        target = OPCM,
        allowFailure = false,
        calldata = abi.encodeCall(upgrade, ())
    ),
    Call3(
        target = Multicall,
        allowFailure = false,
        calldata = abi.encodeCall(Multicall.aggregate3Value, (
            Call3Value(ContractC, false, 100, abi.encodeCall(c, ())),
            Call3Value(ContractD, false, 200, abi.encodeCall(d, ()))
        ))
    )
)
```

### DIFF
--- a/script/universal/MultisigScript.sol
+++ b/script/universal/MultisigScript.sol
@@ -237,10 +237,13 @@ abstract contract MultisigScript is Script {
         }
 
         Call[] memory callsChain = _buildCallsChain({safes: safes});
+
+        vm.startMappingRecording();
         (Vm.AccountAccess[] memory accesses, Simulation.Payload memory simPayload) =
             _simulateForSigner({safes: safes, callsChain: callsChain});
         (StateDiff.MappingParent[] memory parents, string memory json) =
             StateDiff.collectStateDiff(StateDiff.CollectStateDiffOpts({accesses: accesses, simPayload: simPayload}));
+        vm.stopMappingRecording();
 
         _postSign({accesses: accesses, simPayload: simPayload});
         _postCheck({accesses: accesses, simPayload: simPayload});
@@ -513,7 +516,7 @@ abstract contract MultisigScript is Script {
         }
 
         // Return the number of root calls appended.
-        rootCallsCount = rootCallsIndex - rootCallsIndexSaved;
+        return rootCallsIndex - rootCallsIndexSaved;
     }
 
     /// @notice Builds the approve call (`approveHash`) for the given safe and call.
@@ -547,7 +550,7 @@ abstract contract MultisigScript is Script {
     function _printDataToSign(address safe, Call memory call) internal {
         bytes memory txData = _printDataHashes()
             ? _encodeTransactionData({safe: safe, call: call})
-            : _encodeEIP712Json({safe: safe, call: call});
+            : _encodeEip712Json({safe: safe, call: call});
 
         emit DataToSign({data: txData});
 
@@ -748,7 +751,7 @@ abstract contract MultisigScript is Script {
     /// @param call The call to encode.
     ///
     /// @return The EIP-712 JSON structure as bytes.
-    function _encodeEIP712Json(address safe, Call memory call) internal returns (bytes memory) {
+    function _encodeEip712Json(address safe, Call memory call) internal returns (bytes memory) {
         // EIP-712 type definitions for Safe transaction
         string memory types = '{"EIP712Domain":[' '{"name":"chainId","type":"uint256"},'
             '{"name":"verifyingContract","type":"address"}],' '"SafeTx":[' '{"name":"to","type":"address"},'


script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+        return array;
+    }
+
+    /// @notice Wraps the given addresses in an array of two addresses.
+    ///
+    /// @param address1 The first address to wrap.
+    /// @param address2 The second address to wrap.
+    ///
+    /// @return The addresses wrapped in an array of two addresses.
+    function _toArray(address address1, address address2) internal pure returns (address[] memory) {
```
jackchuma: this is unused

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+
+        return call3Values;
+    }
+
+    /// @notice Converts the given calls to the format expected by the `aggregateDelegateCalls` function.
+    ///
+    /// @param calls The calls to get the call3 values for.
+    ///
+    /// @return The calls in the format expected by the `aggregateDelegateCalls` function.
+    function _toDelegateCall3s(Call[] memory calls) internal pure returns (CBMulticall.Call3[] memory) {
```
jackchuma: this is unused in the script

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+
+        return call3s;
+    }
+
+    /// @notice Converts the given calls to the format expected by the `aggregate3Value` function.
+    ///
+    /// @param calls The calls to get the call3 values for.
+    ///
+    /// @return The calls in the format expected by the `aggregate3` function.
+    function _toCall3Values(Call[] memory calls) internal pure returns (CBMulticall.Call3Value[] memory) {
```
jackchuma: this is unused in the script

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+    ///
+    /// @param call The call to get the type for.
+    ///
+    /// @return The type for the given call.
+    function _getCall3Type(Call memory call) internal pure returns (Call3Type) {
+        if (call.operation == Enum.Operation.DelegateCall) {
+            return Call3Type.DELEGATE_CALL;
+        } else if (call.value == 0) {
+            return Call3Type.CALL;
+        } else {
```
jackchuma: Nit: this doesn't need to be in an `else` block

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+                    call.operation, // operation
+                    0, // safeTxGas
+                    0, // baseGas
+                    0, // gasPrice
+                    address(0), // gasToken
+                    payable(address(0)), // refundReceiver
+                    signatures // signatures
+                )
+            ),
+            value: 0
```
jackchuma: is value alwayas 0?

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+    /// @param signatures The signatures to use for the transaction.
+    ///
+    /// @return The execTransaction call.
+    function _buildExecTransactionCall(address safe, Call memory call, bytes memory signatures)
+        internal
+        pure
+        returns (Call memory)
+    {
+        return Call({
+            operation: Enum.Operation.Call,
```
jackchuma: Is the operation always a call here no matter what?

script/universal/MultisigScriptV2.sol
jackchuma, line 1: Nit: please run `forge fmt`

test/universal/MultisigScriptV2.t.sol
```diff
@@ -0,0 +1,196 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {IMulticall3} from "forge-std/interfaces/IMulticall3.sol";
+import {Test} from "forge-std/Test.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {console} from "forge-std/console.sol";
```
jackchuma: Unused import

test/universal/MultisigScriptV2.t.sol
```diff
@@ -0,0 +1,196 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import {IMulticall3} from "forge-std/interfaces/IMulticall3.sol";
```
jackchuma: Unused import

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+// solhint-disable no-console
+import {console} from "lib/forge-std/src/console.sol";
+import {IMulticall3} from "lib/forge-std/src/interfaces/IMulticall3.sol";
```
jackchuma: Unused import

script/universal/MultisigScriptV2.sol
```diff
@@ -0,0 +1,939 @@
...
+    /// Generate a transaction approval data to sign. This method should be called by a threshold of
+    /// multisig owners.
+    ///
+    /// For non-nested multisigs, the signatures can then be used to execute the transaction (see step 3).
+    ///
+    /// For nested multisigs, the signatures can be used to execute an approval transaction for each
+    /// multisig (see step 2).
+    ///
+    /// @param safes A list of nested safes (excluding the executing safe returned by `_ownerSafe`).
+    function sign(address[] memory safes) public {
```
jackchuma: Please add the state diff recording logic back in - removing it breaks the signer tool

script/universal/MultisigScriptV2.sol
xenoliss: Yes but I figured it might still be useful if a script needs to do it (unlikely but it might be the case that we want to build a call to a Multicall from within the script directly). I don't feel strongly here though and ok to remove.

script/universal/MultisigScriptV2.sol
xenoliss: I think the value would/should already be on the safe

script/universal/MultisigScriptV2.sol
xenoliss: I think so. I can't think of a use case where we'd want to delegate call into a safe to call `execTransaction()`.

script/universal/MultisigScriptV2.sol
xenoliss: Yes but I figured it might still be useful if a script needs to do it (unlikely but it might be the case that we want to build a call to a Multicall from within the script directly). I don't feel strongly here though and ok to remove.

foundry.toml
```diff
@@ -6,4 +6,7 @@ optimizer_runs = 999999
 solc_version = "0.8.15"
 extra_output_files = ["abi"]
 
-# See more config options https://github.com/foundry-rs/foundry/tree/master/config
\ No newline at end of file
+[lint]
+lint_on_build = false
```
jackchuma, line 10: Not overly opinionated on this but I'd prefer to leave the linter enabled. I've found it helpful

script/universal/MultisigScript.sol
```diff
@@ -230,14 +236,11 @@ abstract contract MultisigScript is Script {
             originalNonces[i] = _getNonce({safe: safes[i]});
         }
 
-        (bytes[] memory datas, uint256 value) = _transactionDatas({safes: safes});
-
-        vm.startMappingRecording();
```
jackchuma, line 235: Removing `vm.startMappingRecording()` will break some functionality in the signer tool

script/universal/MultisigScript.sol
```diff
@@ -341,72 +350,209 @@ abstract contract MultisigScript is Script {
...
+                });
+            }
+
+            rootCalls[rootCallsIndex] = rootCall;
+            rootCallsIndex++;
         }
-        return dCalls;
+
+        // Return the number of root calls appended.
+        rootCallsCount = rootCallsIndex - rootCallsIndexSaved;
```
jackchuma: It goes against our style guide to not explicitly return here

script/universal/MultisigScript.sol
```diff
@@ -423,80 +569,91 @@ abstract contract MultisigScript is Script {
...
-                    data: datas[i],
-                    value: value,
-                    signatures: Signatures.genPrevalidatedSignature(signer)
-                })
+            execTransactionCalls[i] = _buildExecTransactionCall({
+                safe: safes[i], call: callsChain[i], signatures: Signatures.genPrevalidatedSignature(signer)
             });
         }
-
-        return calls;
```
jackchuma, line 499: Style guide returns

script/universal/MultisigScript.sol
```diff
@@ -593,30 +742,130 @@ abstract contract MultisigScript is Script {
...
     }
 
-    function _execTransactionCalldata(address safe, bytes memory data, uint256 value, bytes memory signatures)
+    /// @notice Encodes the transaction as EIP-712 structured JSON for hardware wallet signing.
+    ///
+    /// @param safe The address of the safe that will execute the transaction.
+    /// @param call The call to encode.
+    ///
+    /// @return The EIP-712 JSON structure as bytes.
+    function _encodeEIP712Json(address safe, Call memory call) internal returns (bytes memory) {
```
jackchuma: Should we use mixed case to adhere to standard solidity style guides here?

foundry.toml
xenoliss: This is a temporary workaround for this issue https://github.com/foundry-rs/foundry/issues/11668

script/universal/MultisigScript.sol
xenoliss: Not too opinionated but I think implicit return makes sense here as we're building the array element by element (and the returned value is named)

jackchuma (approved):

PR #172, opened 2026-01-05
Add MultisigScriptDeposit extension for L1→L2 deposits

leopoldjoy:
This PR adds `MultisigScriptDeposit`, an extension of `MultisigScript` that simplifies L1→L2 deposit transactions.

More specifically:
- Task writers define L2 calls via `_buildL2Calls()` instead of manually wrapping in `depositTransaction`
- Automatically wraps L2 calls in `OptimismPortal.depositTransaction` targeting `CBMulticall` on L2
- Supports ETH bridging (sums values from L2 calls)

### DIFF
--- a/script/universal/MultisigScriptDeposit.sol
+++ b/script/universal/MultisigScriptDeposit.sol
@@ -132,6 +132,7 @@ abstract contract MultisigScriptDeposit is MultisigScript {
     ///      specified `value` field - no additional developer action is required.
     function _buildCalls() internal view virtual override returns (Call[] memory) {
         CBMulticall.Call3Value[] memory l2Calls = _buildL2Calls();
+        require(l2Calls.length > 0, "MultisigScriptDeposit: no L2 calls");
         uint256 totalValue = _sumL2CallValues(l2Calls);
 
         // Encode L2 calls as a multicall


script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,153 @@
...
+///         L2 multicall. A future enhancement could allow each L2 call to specify its own gas
+///         requirement, with the total being calculated automatically.
+abstract contract MultisigScriptDeposit is MultisigScript {
+    //////////////////////////////////////////////////////////////////////////////////////
+    ///                               Virtual Functions                                ///
+    //////////////////////////////////////////////////////////////////////////////////////
+
+    /// @notice Returns the OptimismPortal address on L1
+    /// @dev This is the portal contract that will be called to initiate the L2 deposit
+    function _optimismPortal() internal view virtual returns (address);
```
jackchuma: I think we can provide a `virtual` function implementation for this below that returns the associated address based on the chain ID. This only exists on Sepolia and Mainnet and to my knowledge the addresses have never changed. Would be nice to not require every task to override this

script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,153 @@
...
+    ///      If the gas limit is too low, the L2 transaction will fail but the deposit
+    ///      will still be recorded (ETH will be stuck until manually recovered).
+    ///
+    ///      Common starting points:
+    ///      - Single simple call: 100,000 - 200,000
+    ///      - Multiple calls or complex operations: 500,000+
+    ///
+    ///      Future enhancement: This could be expanded to calculate gas automatically
+    ///      from per-call gas estimates provided in an extended Call3Value struct.
+    function _l2GasLimit() internal view virtual returns (uint64);
```
jackchuma: It would also be ideal to not have to supply a hardcoded value for this. Is there a way we can simulate the call on a fork below within `_buildCalls` so we can estimate the value during script execution?

script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,300 @@
...
+///      1. L2 Post-Check Hook: Currently, `_postCheck` runs on L1 and cannot verify L2 state changes.
+///         A future enhancement could add an `_postCheckL2` hook that forks L2 state and simulates
+///         the deposit transaction's effect. This is non-trivial because deposit transactions are
+///         not immediately reflected on L2.
+///
+///      2. Per-Call Gas Hints: With automatic gas estimation via L2 fork, per-call gas hints are
+///         less critical. However, they could still be useful when L2 RPC is unavailable or when
+///         task writers want manual control without overriding `_l2GasLimit()` entirely.
+///
+/// @dev Testing Note:
```
jackchuma: Nit: a comment about unit tests could be moved to the test file

jackchuma (commented):
I'm no longer convinced this is the right way to do things. If the calculated gas limit changes between sign, approve and run calls, the signatures will no longer be valid. I think moving this estimation into the signer tool during the validation file generation might make more sense to guarantee that the value does not change. The validation file could get an optional `l2GasLimit` field where, if set, the signer tool knows to add that as an env var. From the context of this script, expecting the env var to be set would work in that case

script/universal/MultisigScriptDeposit.sol
leopoldjoy: Removed now 👍

leopoldjoy:
> I'm no longer convinced this is the right way to do things. If the calculated gas limit changes between sign, approve and run calls, the signatures will no longer be valid. I think moving this estimation into the signer tool during the validation file generation might make more sense to guarantee that the value does not change. The validation file could get an optional `l2GasLimit` field where, if set, the signer tool knows to add that as an env var. From the context of this script, expecting the env var to be set would work in that case

Thanks, good call with this, was an implementation oversight on my part! I've switched back to the env var approach now.

script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    //////////////////////////////////////////////////////////////////////////////////////
+    ///                              Internal Functions                                ///
+    //////////////////////////////////////////////////////////////////////////////////////
+
+    /// @notice Sums the ETH values from an array of L2 calls
+    /// @param l2Calls The array of L2 calls to sum values from
+    /// @return total The total ETH value across all calls
+    function _sumL2CallValues(CBMulticall.Call3Value[] memory l2Calls) internal pure returns (uint256 total) {
```
jackchuma, line 170: Nit: prefer an explicit return here

script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,174 @@
...
+        Call[] memory l1Calls = new Call[](1);
+        l1Calls[0] = Call({
+            operation: Enum.Operation.Call,
+            target: _optimismPortal(),
+            data: abi.encodeCall(
+                IOptimismPortal2.depositTransaction,
+                (
+                    CB_MULTICALL, // L2 target: CBMulticall at same address on L2
+                    totalValue, // ETH to bridge
+                    _l2GasLimit(), // Gas limit for L2 execution
```
jackchuma, line 152: We could potentially consider enforcing a minimum gas limit to protect against an obvious error (like setting a gas limit of 0)

jackchuma (dismissed):

script/universal/MultisigScriptDeposit.sol
leopoldjoy: Yes I thought about this, obviously 0 would be wrong, but there are larger small values which would also very likely be an error, so it felt arbitrary to check only 0, but I can add that if you think its stronger 👍

script/universal/MultisigScriptDeposit.sol
leopoldjoy: Done: https://github.com/base/contracts/pull/172/commits/6890a7bdddeacd27cfa2e8b2264b7f4d0445e956

jackchuma (dismissed):

script/universal/MultisigScriptDeposit.sol
```diff
@@ -0,0 +1,175 @@
...
+    ///      The L2 calls are encoded as a CBMulticall.aggregate3Value call, which is then
+    ///      passed as the data payload to OptimismPortal.depositTransaction. This allows
+    ///      multiple L2 operations to be batched into a single deposit transaction.
+    ///
+    ///      ETH bridging: If any L2 calls include a non-zero `value`, the total ETH is
+    ///      summed and sent with the deposit transaction. The CBMulticall.aggregate3Value
+    ///      function on L2 automatically distributes the ETH to each call according to its
+    ///      specified `value` field - no additional developer action is required.
+    function _buildCalls() internal view virtual override returns (Call[] memory) {
+        CBMulticall.Call3Value[] memory l2Calls = _buildL2Calls();
```
emrahCoinbase, line 134: Would recommend adding length validation. If _buildL2Calls() returns an empty array, the deposit transaction will still be created but with zero calls and zero value. However, this results in a wasted L1 transaction.

PR #196, opened 2026-02-28
feat(CHAIN-3293) Migrate Multiproof Contracts to base/contracts

jjtny1:
This PR migrates the contracts from https://github.com/base/multiproof-dispute-game to base contracts.

Multiproofs depend on a modified version of the DisputeGameFactory that takes arguments in the `initialize` constructor. This is done through a `git patch` found in `multiproof_tests/optimism.patch`. Because this patch is needed for the tests to pass, we keep all multiproof tests in the the `multiproof_tests` folder and use a foundry profile to trigger it. This is abstracted into a make command: `make test-multiproof`


.github/workflows/test.yml
```diff
@@ -60,5 +60,5 @@ jobs:
           git diff --exit-code snapshots/semver-lock.json
 
       - name: Run Forge tests
-        run: just test
+        run: just test && just test-multiproof
```
jjtny1: Temporary and should be removed once disputegame changes are made

test/multiproof/AggregateVerifier.t.sol
jackchuma, line 1: Could this be a `multiproof` directory inside of the existing `test/` directory?

deploy-config/sepolia-no-nitro.json
jackchuma, line 1: Not blocking for this PR, but these deploy config formats do not match the existing deploy configs in this repo

multiproof_tests/optimism.patch
jackchuma, line 1: We no longer need patch files - can just update the contracts. This can be a follow up PR though

src/multiproof/mocks/MockDevSystemConfigGlobal.sol
jackchuma, line 1: Nit: this should be moved to a mocks folder somewhere (can be done in follow up PR)

src/multiproof/tee/SystemConfigGlobal.sol
```diff
@@ -0,0 +1,144 @@
...
+import { LibBytes } from "@nitro-validator/LibBytes.sol";
+import { NitroValidator } from "@nitro-validator/NitroValidator.sol";
+import { OwnableManagedUpgradeable } from "@op-enclave/OwnableManagedUpgradeable.sol";
+
+/// @title SystemConfigGlobal
+/// @notice Manages TEE signer registration via AWS Nitro attestation.
+/// @dev Signers are registered by providing a valid AWS Nitro attestation document.
+///      Each signer is associated with the PCR0 (enclave image hash) from their attestation,
+///      which allows TEEVerifier to validate that a signer was registered with a specific image.
+contract SystemConfigGlobal is OwnableManagedUpgradeable, NitroValidator {
```
jackchuma: Lets add the ISemver interface in a follow up PR

src/multiproof/tee/TEEVerifier.sol
```diff
@@ -0,0 +1,84 @@
...
+/// @title TEEVerifier
+/// @notice Stateless TEE proof verifier that validates signatures against registered signers.
+/// @dev This contract is designed to be used as the TEE_VERIFIER in the AggregateVerifier.
+///      It verifies that proofs are signed by enclave addresses registered in SystemConfigGlobal
+///      via AWS Nitro attestation, and that the signer's PCR0 matches the claimed imageId.
+///      Additionally, it verifies that the L1 origin block referenced in the proof actually exists
+///      by checking against blockhash() or the EIP-2935 history contract.
+///      The contract is intentionally stateless - all state related to output proposals is
+///      managed by the calling contract (e.g., AggregateVerifier).
+contract TEEVerifier is IVerifier {
```
jackchuma: Lets add the ISemver interface in a follow up PR

src/multiproof/AggregateVerifier.sol
```diff
@@ -0,0 +1,917 @@
...
+import { Claim, GameStatus, GameType, Hash, Proposal, Timestamp } from "src/dispute/lib/Types.sol";
+
+// Solady
+import { Clone } from "@solady/utils/Clone.sol";
+import { FixedPointMathLib } from "@solady/utils/FixedPointMathLib.sol";
+import { ReentrancyGuard } from "@solady/utils/ReentrancyGuard.sol";
+
+import { IVerifier } from "interfaces/multiproof/IVerifier.sol";
+
+contract AggregateVerifier is Clone, ReentrancyGuard {
```
jackchuma: Let's add the ISemver interface in a follow up PR

foundry.toml
```diff
@@ -43,13 +43,16 @@ remappings = [
...
+  'solady/=lib/solady/src/',
   '@solady/=lib/solady/src',
   '@solady-v0.0.245/=lib/solady-v0.0.245/src',
   'forge-std/=lib/forge-std/src',
   'ds-test/=lib/forge-std/lib/ds-test/src',
   'safe-contracts/=lib/safe-contracts/contracts',
   'kontrol-cheatcodes/=lib/kontrol-cheatcodes/src',
-  'interfaces/=interfaces'
+  'interfaces/=interfaces',
+  '@nitro-validator/=lib/nitro-validator/src/',
```
jackchuma: I'd strongly prefer not to add more remappings. It just makes this repo harder to work with from contract-deployments

foundry.toml
```diff
@@ -173,3 +176,12 @@ src = 'test/kontrol/proofs'
 out = 'kout-proofs'
 test = 'test/kontrol/proofs'
 script = 'test/kontrol/proofs'
+
+################################################################
+#                         PROFILE: Multiproof                  #
+################################################################
+
+[profile.multiproof]
+src = 'multiproof_tests'
```
jackchuma: This doesn't make sense to me - do we need a multiproof profile?

justfile
```diff
@@ -79,6 +79,15 @@ test *ARGS: build-go-ffi
 test-dev *ARGS: build-go-ffi
   FOUNDRY_PROFILE=lite forge test {{ARGS}}
 
+# Runs multiproof contract tests.
+test-multiproof *ARGS: build-go-ffi
```
jackchuma: Is this still necessary if our tests get moved to the existing test directory? It should be removed if possible

Makefile
```diff
@@ -44,3 +56,15 @@ bindings:
 	abigen --abi out/BalanceTracker.sol/BalanceTracker.abi.json --pkg bindings --type BalanceTracker --out bindings/balance_tracker.go
 	abigen --abi out/FeeDisburser.sol/FeeDisburser.abi.json --pkg bindings --type FeeDisburser --out bindings/fee_disburser.go
 	cd bindings && go mod tidy
+
+.PHONY: checkout-op-commit
```
jackchuma: Remove this

Makefile
```diff
@@ -32,6 +34,16 @@ deps: clean-lib
 test:
 	forge test --ffi -vvv
 
+.PHONY: test-multiproof
```
jackchuma: Remove this if possible. We don't need it in both the makefile and justfile anyways

.gitignore
```diff
@@ -7,6 +7,7 @@ kout-proofs
 test/kontrol/logs
 out
 lib
+multiproof-out
```
jackchuma: We shouldn't need this

test/vendor/Initializable.t.sol
```diff
@@ -395,6 +395,11 @@ contract Initializer_Test is CommonTest {
         // Contract is not deployed as part of the standard deployment script.
         excludes[j++] = "src/revenue-share/BalanceTracker.sol";
 
+        // Exclude Multiproof contracts as they are not part of standard deployment script.
```
jackchuma: They should be added to the standard deployment script

test/libraries/SemverComp.t.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity 0.8.15;
 import { Test } from "forge-std/Test.sol";
 
 // Libraries
-import { JSONParserLib } from "solady/src/utils/JSONParserLib.sol";
+import { JSONParserLib } from "@solady/utils/JSONParserLib.sol";
```
jackchuma, line 8: Was this change necessary? The import worked before

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Yes. Also, note that this file will be significantly overhauled by my ZK verifier rework PR which I'll open shortly. (It switches to ZK verifying via the Automata SDK instead of the existing op-enclave-style onchain certificate attestation approach which is currently used here.)

src/multiproof/tee/TEEVerifier.sol
leopoldjoy: Sure, I'll add it in my PR which overhauls the `SystemConfigGlobal.sol` contract.

src/multiproof/AggregateVerifier.sol
leopoldjoy: I'll add 👍

test/multiproof/AggregateVerifier.t.sol
jjtny1: I originally had it like that but I couldnt figure out a way to tell `forge test` to skip the multiproof directory. The multiproof tests need the `optimism.patch` applied or else they fail and the other tests fail when that patch is applied so i separated them

multiproof_tests/optimism.patch
jjtny1: Oh i wasnt sure if we could update the contracts directly. ill make this change and that solves the testing issue as well

src/multiproof/mocks/MockDevSystemConfigGlobal.sol
jjtny1: Ill move it 👍

foundry.toml
jjtny1: okay @leopoldjoy what do you think i can change to using `lib/nitro-validator`

foundry.toml
jjtny1: This was to allow testing multiproof separately from the regular tests because of the patch. But seems like we dont need the patch so dont need this anymore

justfile
jjtny1: Yeah this was only needed because of the patch

test/libraries/SemverComp.t.sol
jjtny1: I keep getting this error:

```
2026-03-01T00:22:54.808255Z ERROR foundry_compilers_artifacts_solc::sources: error="/Users/jobythundil/Documents/coinbase_dev/go/src/github.com/base/multiproof-dispute-game-2/lib/solady/src/src/utils/JSONParserLib.sol": No such file or directory (os error 2)
[⠃] Compiling...2026-03-01T00:22:54.904261Z ERROR foundry_compilers_artifacts_solc::sources: error="/Users/jobythundil/Documents/coinbase_dev/go/src/github.com/base/multiproof-dispute-game-2/lib/solady/src/src/utils/JSONParserLib.sol": No such file or directory (os error 2)

```

jjtny1: This could be an issue with my local cache?

jackchuma: Strange, ok maybe something else about this PR is causing it. I'm ok with it then

deploy-config/sepolia-no-nitro.json
leopoldjoy: @jjtny1 no worries about changing this here, I've fixed it here: https://github.com/base/contracts/pull/197

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: I've done this here: https://github.com/base/contracts/pull/197

src/multiproof/tee/TEEVerifier.sol
leopoldjoy: https://github.com/base/contracts/pull/197

src/multiproof/AggregateVerifier.sol
leopoldjoy: https://github.com/base/contracts/pull/197

foundry.toml
leopoldjoy: No worries about it, I removed the mapping and switched all references in this PR: https://github.com/base/contracts/pull/197

test/vendor/Initializable.t.sol
leopoldjoy: I've removed the unnecessary ones here: https://github.com/base/contracts/pull/197

leopoldjoy (approved):

scripts/autogen/generate-semver-lock/main.go
```diff
@@ -81,6 +81,16 @@ func processFile(file string) (*SemverLockResult, []error) {
 		sourceFilePath = path
 		contractName = name
 		contractKey = sourceFilePath + ":" + name
+		if strings.HasSuffix(file, ".dispute.json") {
```
jackchuma, line 84: Do we need this?

snapshots/abi/MockSystemConfig.json
jackchuma, line 1: I'm not sure we need mock abis in our snapshots

snapshots/abi/MockVerifier.json
jackchuma, line 1: Don't need this

snapshots/abi/DevSystemConfigGlobal.json
jackchuma, line 1: Dont need this

snapshots/storageLayout/DevSystemConfigGlobal.json
jackchuma, line 1: Don't need this

snapshots/storageLayout/MockSystemConfig.json
jackchuma, line 1: Don't need this

snapshots/storageLayout/MockVerifier.json
jackchuma, line 1: Don't need this

src/dispute/zk/OPSuccinctFaultDisputeGame.sol
```diff
@@ -321,6 +321,8 @@ contract OPSuccinctFaultDisputeGame is Clone, ISemver, IDisputeGame {
             GameType.unwrap(ANCHOR_STATE_REGISTRY.respectedGameType()) == GameType.unwrap(GAME_TYPE);
     }
 
+    function initializeWithInitData(bytes calldata) external payable { }
```
jackchuma, line 324: Why is this here if it's empty?

foundry.toml
```diff
@@ -21,6 +21,7 @@ additional_compiler_profiles = [
   { name = "dispute", optimizer_runs = 5000 },
 ]
 compilation_restrictions = [
```
jackchuma, line 23: Why are a bunch of compilation restrictions changed for contracts that are not relevant to multi proofs?

foundry.toml
```diff
@@ -153,6 +163,7 @@ additional_compiler_profiles = [
   { name = "dispute", optimizer_runs = 0 },
 ]
 compilation_restrictions = [
```
jackchuma, line 165: same here regarding compilation restriction changes

scripts/autogen/generate-semver-lock/main.go
jjtny1: Yes, the reason is because `forge build` produces an artifact _per compiler profile_ ([code](https://github.com/base/contracts/blob/main/foundry.toml#L21)) for some contracts.

This results in a non-deterministic generation of the semver.json file. When searching through `forge-artifacts` sometimes the default compiler profile is picked up and sometimes it's the dispute profile optimization. The [contract key](https://github.com/base/contracts/blob/main/scripts/autogen/generate-semver-lock/main.go#L83) doesn't differentiate between these profiles so on different runs it produces different hashes.

This causes the CI tests to be non-deterministic.

This fixes the issue by checking if we are using the dispute optimized code and adds that to the contract key stored in the json

src/dispute/zk/OPSuccinctFaultDisputeGame.sol
jjtny1: It needs to be implemented since this contract inherits IDisputeGame and we added this function for the AggregateVerifier

jjtny1: It needs to be implemented since this contract inherits IDisputeGame and we added this function for the AggregateVerifier

foundry.toml
jjtny1: Yeah this was the tricky part with getting CI to pass.

Our multiproof contracts, directly import (or indirectly) import these contracts. This causes the forge compiler to produce 2 artifacts (one per[ compiler profile](https://github.com/base/contracts/blob/main/foundry.toml#L21)). 

Having more than 1 artifact causes the following error:
1. semver is non-deterministic (see comment above)
2. vm.getCode (or vm.getDeploymentCode) cheatcodes cause tests to fail because they pick up different artifacts. An example of this is the VerifyOPCM tests. When the OP contracts are deployed, they are deployed with the code from default profile. When the tests are run, the call to vm.GetCode picks up the `dispute` optimized profile. This results in 2 different bytecodes so [the bytecode comparison check fails](https://github.com/base/contracts/blob/be7c7a642e430fa64b04b63203839f8c81f48466/scripts/deploy/VerifyOPCM.s.sol#L471).

The fundamental issue here is that the `dispute`compiler profile causes forge to compile multiple artifacts for a single contract (one optimized the other not). From what I can tell, the reason this error shows up now is because our multiproof contracts somehow trigger different compilations per profile so more than one artifact is created.

jjtny1: Im not sure if there is a way for us to tell forge "only use the default compiler profile or optimized one". By adding these contracts, we force these contracts to only use the optimized version

foundry.toml
jjtny1: see comment above

src/dispute/zk/OPSuccinctFaultDisputeGame.sol
jackchuma: Should it be empty though?

jjtny1: It shouldnt be used since init extra data is something we introduced -- everything using it currently should only be using the initialize function and this contract has no need of the extra data

PR #197, opened 2026-03-02
Replace onchain Nitro cert verification with Automata ZK verifier

leopoldjoy:
This PR:
- Replaces `NitroValidator/CertManager` with external `NitroEnclaveVerifier` (Risc0 ZK path) in `SystemConfigGlobal`.
- Adds `ISemver` to `SystemConfigGlobal`, `TEEVerifier`, and `AggregateVerifier`.
- Wires `SystemConfigGlobal` into standard `Deploy.s.sol` pipeline and `Initializable.t.sol`.
- Integrates multiproof config into standard `DeployConfig.s.sol` and migrates dev deploy scripts to use it, removing separate `sepolia-no-nitro.json` / `sepolia-with-nitro.json` config files.
- Fixes `TEEVerifier` proof format to match `AggregateVerifier`'s `l1head` change (85 bytes instead of 149).
- Adds `pubKey` length check in `SystemConfigGlobal.registerSigner` for defense-in-depth.

### DIFF
--- a/snapshots/semver-lock.json
+++ b/snapshots/semver-lock.json
@@ -249,11 +249,11 @@
   },
   "src/multiproof/tee/SystemConfigGlobal.sol:SystemConfigGlobal": {
     "initCodeHash": "0x76da4f2a736d7a39a01720e5d900a85fcaa60ba0430fcacbb8ab367f55ba5411",
-    "sourceCodeHash": "0xba78bc35f234479e9844b595f109da6029d46c062d246af6dc410542c84c61a3"
+    "sourceCodeHash": "0xa6261402efe0105e2a4f9369818bafb4e65515e51850b44d47504151e1c39d01"
   },
   "src/multiproof/tee/SystemConfigGlobal.sol:SystemConfigGlobal:dispute": {
     "initCodeHash": "0xfae3a71157f3c64a7bda037ec116bf6e7265099397d9bcad22c01cc1f029ed7d",
-    "sourceCodeHash": "0xba78bc35f234479e9844b595f109da6029d46c062d246af6dc410542c84c61a3"
+    "sourceCodeHash": "0xa6261402efe0105e2a4f9369818bafb4e65515e51850b44d47504151e1c39d01"
   },
   "src/multiproof/tee/TEEVerifier.sol:TEEVerifier": {
     "initCodeHash": "0x78317d9088a26523d938ce0d88ed0134abc60292c0cdb3f8a6a9530638fd9e9a",


justfile
```diff
@@ -75,6 +75,10 @@ clean:
 test *ARGS: build-go-ffi
   forge test {{ARGS}}
 
+# Runs multiproof contract tests.
+test-multiproof *ARGS: build-go-ffi
```
jackchuma: Why do we need a test command for multiproofs. Why can't we just use the existing test command?

src/multiproof/tee/SystemConfigGlobal.sol
```diff
@@ -88,27 +100,33 @@ contract SystemConfigGlobal is OwnableManagedUpgradeable, NitroValidator {
...
-        if (ptrs.timestamp / MS_PER_SECOND + MAX_AGE <= block.timestamp) revert AttestationTooOld();
 
-        // The publicKey is encoded in the form specified in section 4.3.6 of ANSI X9.62,
-        // which is a 0x04 byte followed by the x and y coordinates of the public key.
-        // We ignore the first byte when hashing.
-        bytes32 publicKeyHash = attestationTbs.keccak(ptrs.publicKey.start() + 1, ptrs.publicKey.length() - 1);
+        // The publicKey is encoded in ANSI X9.62 format: 0x04 || x || y (65 bytes).
+        // We skip the first byte (0x04 prefix) when hashing to derive the address.
+        bytes memory pubKey = journal.publicKey;
+        if (pubKey.length < 2) revert InvalidPublicKey();
```
jackchuma: Since pubkeys are expected to be 65 bytes, should this check be `if (pubKey.length != 65) ...` ?

src/multiproof/tee/SystemConfigGlobal.sol
```diff
@@ -137,8 +155,15 @@ contract SystemConfigGlobal is OwnableManagedUpgradeable, NitroValidator {
...
-    /// @custom:semver 0.1.0
+    /// @custom:semver 0.2.0
     function version() public pure virtual returns (string memory) {
-        return "0.1.0";
+        return "0.2.0";
+    }
+
+    /// @dev Extracts PCR0 from the first entry in the PCR array and returns its keccak256 hash.
+    function _extractPCR0Hash(Pcr[] memory pcrs) internal pure returns (bytes32) {
+        if (pcrs.length == 0 || pcrs[0].index != 0) revert PCR0NotFound();
```
jackchuma: Should we iterate over `pcrs` to find `index == 0` or is it safe to assume this will always be the first element?

scripts/deploy/DeployImplementations.s.sol
```diff
@@ -704,27 +711,25 @@ contract DeployImplementations is Script {
...
-    function deployAggregateVerifierImpl(Input memory, Output memory _output) private {
-        DeployDevWithNitro nitro = new DeployDevWithNitro();
-        DeployDevWithNitro.DeployConfig memory cfg = nitro.loadConfig();
-
+    function deployAggregateVerifierImpl(Input memory _input, Output memory _output) private {
         address zkVerifier = address(new MockVerifier());
 
-        address certManager = address(new CertManager());
-        SystemConfigGlobal scgImpl = new SystemConfigGlobal(CertManager(certManager));
+        SystemConfigGlobal scgImpl = new SystemConfigGlobal(INitroEnclaveVerifier(address(0)));
```
jackchuma: Do we have a nitro enclave verifier address that can be used here? Maybe it should be added to the `Input` struct instead of hardcoding to use address 0?

scripts/multiproof/DeployDevNoNitro.s.sol
```diff
@@ -172,65 +146,64 @@ contract DeployDevNoNitro is Script {
...
 
-        // 5. Mock AnchorStateRegistry
         MockAnchorStateRegistry asr = new MockAnchorStateRegistry();
         mockAnchorRegistry = address(asr);
-        asr.initialize(disputeGameFactory, Hash.wrap(cfg.genesisOutputRoot), cfg.genesisBlockNumber, cfg.gameType);
+        asr.initialize(
+            disputeGameFactory,
+            Hash.wrap(cfg.multiproofGenesisOutputRoot()),
+            cfg.multiproofGenesisBlockNumber(),
+            GameType.wrap(uint32(cfg.multiproofGameType()))
```
jackchuma: Nit: this `GameType.wrap(uint32(cfg.multiproofGameType()))` expression is repeated a few times, it could be a local variable

deploy-config/sepolia.json
```diff
@@ -1,5 +1,5 @@
 {
-  "finalSystemOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "finalSystemOwner": "0x6e427c3212C0b63BE0C382F97715D49b011bFF33",
```
jackchuma: What's this address?

deploy-config/sepolia.json
```diff
@@ -64,5 +64,12 @@
   "gasPayingTokenName": "",
   "gasPayingTokenSymbol": "",
   "nativeAssetLiquidityAmount": null,
-  "liquidityControllerOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301"
+  "liquidityControllerOwner": "0xfd1D2e729aE8eEe2E146c033bf4400fE75284301",
+  "teeImageHash": "0x0000000000000000000000000000000000000000000000000000000000000001",
+  "multiproofConfigHash": "0x12e9c45f19f9817c6d4385fad29e7a70c355502cf0883e76a9a7e478a85d1360",
+  "multiproofGameType": 621,
+  "teeProposer": "0xb28E6890Cffa969dA9851c1BF1Ac34B76EbFEe98",
+  "nitroEnclaveVerifier": "0x0000000000000000000000000000000000000000",
```
jackchuma, line 72: Do we have an address we can use for this?

deploy-config/hardhat.json
```diff
@@ -72,5 +72,12 @@
   "gasPayingTokenName": "",
   "gasPayingTokenSymbol": "",
   "nativeAssetLiquidityAmount": null,
-  "liquidityControllerOwner": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc"
+  "liquidityControllerOwner": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+  "teeImageHash": "0x0000000000000000000000000000000000000000000000000000000000000001",
+  "multiproofConfigHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "multiproofGameType": 621,
+  "teeProposer": "0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc",
+  "nitroEnclaveVerifier": "0x0000000000000000000000000000000000000000",
```
jackchuma, line 80: Do we have an address we can use for this?

justfile
leopoldjoy: Yeah, this was to fix a CI build failure, but good call. I resolved the root problem in the `test.yml` file here: https://github.com/base/contracts/pull/197/commits/3bf1d92cfef10d8bb93b1eae3e59a2119e0ebbf6.

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Good catch, fixed: https://github.com/base/contracts/pull/197/commits/3bf1d92cfef10d8bb93b1eae3e59a2119e0ebbf6

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Technically it should be in the first position, as that's the standard for AWS. However, upon reflection following your comment, I agree that using a loop would increase the resilience (maybe there's some edge case where something's broken in AWS, etc.). Changed [here](https://github.com/base/contracts/pull/197/commits/3bf1d92cfef10d8bb93b1eae3e59a2119e0ebbf6).

scripts/deploy/DeployImplementations.s.sol
leopoldjoy: We don't yet, but this should be read in anyway, [fixed](https://github.com/base/contracts/pull/197/commits/3bf1d92cfef10d8bb93b1eae3e59a2119e0ebbf6).

deploy-config/sepolia.json
leopoldjoy: Reset: https://github.com/base/contracts/pull/197/commits/3bf1d92cfef10d8bb93b1eae3e59a2119e0ebbf6

deploy-config/sepolia.json
leopoldjoy: No, not yet. The `NitroEnclaveVerifier` hasn't been deployed. The `address(0)` is just a placeholder for now.

deploy-config/hardhat.json
leopoldjoy: Same. Not yet. The `NitroEnclaveVerifier` hasn't been deployed. `The address(0)` is just a placeholder for now.

src/multiproof/tee/SystemConfigGlobal.sol
```diff
@@ -137,8 +155,19 @@ contract SystemConfigGlobal is OwnableManagedUpgradeable, NitroValidator {
     }
 
     /// @notice Semantic version.
-    /// @custom:semver 0.1.0
+    /// @custom:semver 0.2.0
```
jackchuma: Do we need to increment semver? This hasn't been deployed yet

src/multiproof/tee/SystemConfigGlobal.sol
```diff
@@ -88,27 +100,33 @@ contract SystemConfigGlobal is OwnableManagedUpgradeable, NitroValidator {
...
-        Ptrs memory ptrs = validateAttestation(attestationTbs, signature);
-        bytes32 pcr0Hash = attestationTbs.keccak(ptrs.pcrs[0]);
+    /// @notice Registers a signer using a ZK proof of an AWS Nitro attestation document.
+    /// @dev The ZK proof must verify a valid attestation that:
+    ///      1. Has a valid AWS Nitro certificate chain (verified offchain via ZK)
+    ///      2. Contains a PCR0 that has been pre-registered via registerPCR0
+    ///      3. Is less than MAX_AGE old
+    /// @param output The ABI-encoded VerifierJournal from the ZK proof.
+    /// @param proofBytes The Risc0 ZK proof bytes.
+    function registerSigner(bytes calldata output, bytes calldata proofBytes) external onlyOwnerOrManager {
```
jackchuma, line 110: We need testing for this fn

src/multiproof/tee/TEEVerifier.sol
jackchuma: This comment is misleading - `TeeVerifier.verify()` does not seem to verify L1 origin. This is handled by `AggregateVerifier` correct?

scripts/deploy/DeployConfig.s.sol
```diff
@@ -192,6 +201,13 @@ contract DeployConfig is Script {
         faultGameV2SplitDepth = _readOr(_json, "$.faultGameV2SplitDepth", 30);
         faultGameV2ClockExtension = _readOr(_json, "$.faultGameV2ClockExtension", 10800);
         faultGameV2MaxClockDuration = _readOr(_json, "$.faultGameV2MaxClockDuration", 302400);
+        teeImageHash = bytes32(_readOr(_json, "$.teeImageHash", 0));
+        multiproofConfigHash = bytes32(_readOr(_json, "$.multiproofConfigHash", 0));
+        multiproofGameType = _readOr(_json, "$.multiproofGameType", 621);
+        teeProposer = _readOr(_json, "$.teeProposer", finalSystemOwner);
+        nitroEnclaveVerifier = _readOr(_json, "$.nitroEnclaveVerifier", address(0));
```
jackchuma: We should not fall back to `address(0)` here - let's instead use `stdJson.readAddress` without a fallback

scripts/deploy/DeployImplementations.s.sol
```diff
@@ -706,32 +712,25 @@ contract DeployImplementations is Script {
...
                     _output.anchorStateRegistryImpl,
                     _output.delayedWETHImpl,
                     IVerifier(teeVerifierImpl),
                     IVerifier(zkVerifier),
-                    cfg.teeImageHash,
+                    _input.teeImageHash,
                     bytes32(0),
-                    cfg.configHash,
+                    _input.multiproofConfigHash,
                     8453,
```
jackchuma: I doubt we want a hardcoded chain ID here - can this be read from the input struct?

jjtny1 (approved):
Aside from Jack's comments, lgtm

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: It's an incrementation based on iterating away from op-enclave's original version. Maybe this doesn't matter though, what do you think?

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Testing for this is quite in-depth since it involves providing a ZK proof, I was planning to add this in a follow-up PR, but let me know if you feel strongly about the need for it here.

src/multiproof/tee/SystemConfigGlobal.sol
jackchuma: ah right, ok I'm fine with the increment then

src/multiproof/tee/SystemConfigGlobal.sol
jackchuma: I feel strongly about the need for it before this goes into audit. I'm fine with it being added in a separate PR though. We could use a mock verifier contract so you don't actually have to provide a zk proof in the test right?

scripts/deploy/DeployImplementations.s.sol
leopoldjoy: Fixed: https://github.com/base/contracts/pull/197/changes/89d24bc4b037fb76658749fb2f2765516a8f96c0

scripts/deploy/DeployConfig.s.sol
leopoldjoy: Fixed: https://github.com/base/contracts/pull/197/changes/89d24bc4b037fb76658749fb2f2765516a8f96c0

src/multiproof/tee/TEEVerifier.sol
leopoldjoy: Good catch, it used to be handled by `TeeVerifier`, I've fixed it here: https://github.com/base/contracts/pull/197/changes/89d24bc4b037fb76658749fb2f2765516a8f96c0

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Actually, I agree with you, since now with this rework to using offchain certificate verification, the contract has changed so meaningfully that it likely shouldn't be considered as progressive iteration from op-enclave, but instead as an entirely new system. I've updated it here: https://github.com/base/contracts/pull/197/commits/054df85beed018d12b564423269743a6183e80c3

src/multiproof/tee/SystemConfigGlobal.sol
leopoldjoy: Yes, 100%. I will follow-up this PR shortly with another PR with that kind of testing setup.

src/multiproof/tee/SystemConfigGlobal.sol
jackchuma: Need to update this comment back to 0.1.0

src/multiproof/tee/TEEVerifier.sol
```diff
@@ -43,15 +42,15 @@ contract TEEVerifier is IVerifier {
...
     /// @notice Verifies a TEE proof for a state transition.
-    /// @param proofBytes The proof: proposer (20) + l1OriginHash (32) + l1OriginNumber (32) + signature (65) = 149
-    /// bytes. @param imageId The claimed TEE image hash (PCR0). Must match the signer's registered PCR0.
+    /// @param proofBytes The proof: proposer(20) + signature(65) = 85 bytes.
+    /// @param imageId The claimed TEE image hash (PCR0). Must match the signer's registered PCR0.
     /// @param journal The keccak256 hash of the proof's public inputs.
     /// @return valid Whether the proof is valid.
     function verify(bytes calldata proofBytes, bytes32 imageId, bytes32 journal) external view override returns (bool) {
-        if (proofBytes.length < 149) revert InvalidProofFormat();
+        if (proofBytes.length < 85) revert InvalidProofFormat();
```
jackchuma, line 50: Nit: 85 and 20 are used as magic numbers in this fn. Could be named constants

leopoldjoy: Good point, I'll include this in a follow-up PR.

PR #198, opened 2026-03-03
feat(CHAIN-3335): add FlashblockIndex contract

xenoliss:
## Summary

Add a minimal `FlashblockIndex` contract that stores the current flashblock index alongside `block.number`. The builder calls this contract at the start of each flashblock via `fallback()` with 1 byte of calldata (the index as `uint8`). Both values are manually packed into a single `uint256` to guarantee exactly 1 SSTORE per write.

This enables on-chain consumers (contracts, searchers) to read which flashblock they are executing in within a block by calling `get()`, which returns `(flashblockIndex, blockNumber)` in a single SLOAD.

## Design decisions

- **`fallback()` instead of a named function**: eliminates the 4-byte selector overhead, reducing DA cost to 1 byte of calldata per flashblock
- **Manual bit-packing**: Manual packing of the flashblock index and block number to compile down to 1 SSTORE/SLOAD
- **`immutable BUILDER`**: stored in bytecode, no storage read for the auth check

### DIFF
--- a/src/L2/FlashblockIndex.sol
+++ b/src/L2/FlashblockIndex.sol
@@ -9,17 +9,6 @@ import { ISemver } from "interfaces/universal/ISemver.sol";
 /// @dev The builder calls this via fallback with 1 byte of calldata (the index as uint8).
 ///      Both values are manually packed into a single uint256 to guarantee 1 SSTORE per write.
 contract FlashblockIndex is ISemver {
-    /// @notice Thrown when the caller is not the authorized builder.
-    error OnlyBuilder();
-
-    /// @notice Thrown when calldata is not exactly 1 byte.
-    error InvalidCalldata();
-
-    /// @notice Emitted when the flashblock index is updated.
-    /// @param flashblockIndex The new flashblock index.
-    /// @param blockNumber The block number at which the index was set.
-    event FlashblockIndexUpdated(uint8 indexed flashblockIndex, uint48 indexed blockNumber);
-
     /// @notice Semantic version.
     /// @custom:semver 1.0.0
     string public constant override version = "1.0.0";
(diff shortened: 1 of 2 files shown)


src/L2/FlashblockIndex.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    /// @notice Constructor.
+    /// @param builder The address authorized to update the flashblock index.
+    constructor(address builder) {
+        BUILDER = builder;
+        _disableInitializers();
+    }
+
+    /// @notice Initializer.
+    function initialize() external initializer { }
```
jackchuma: Why include this if it's empty?

src/L2/FlashblockIndex.sol
```diff
@@ -0,0 +1,63 @@
...
+    error InvalidCalldata();
+
+    /// @notice Emitted when the flashblock index is updated.
+    /// @param flashblockIndex The new flashblock index.
+    /// @param blockNumber The block number at which the index was set.
+    event FlashblockIndexUpdated(uint8 indexed flashblockIndex, uint48 indexed blockNumber);
+
+    /// @notice Semantic version.
+    /// @custom:semver 1.0.0
+    string public constant override version = "1.0.0";
```
jackchuma, line 14: Nit: I think most other contracts define this as an external view function instead of a constant variable

src/L2/FlashblockIndex.sol
```diff
@@ -0,0 +1,63 @@
...
+import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import { ISemver } from "interfaces/universal/ISemver.sol";
+
+/// @custom:upgradeable
+/// @title FlashblockIndex
+/// @notice Stores the current flashblock index alongside block.number.
+/// @dev The builder calls this via fallback with 1 byte of calldata (the index as uint8).
+///      Both values are manually packed into a single uint256 to guarantee 1 SSTORE per write.
+contract FlashblockIndex is Initializable, ISemver {
+    /// @notice Thrown when the caller is not the authorized builder.
```
jackchuma: Nit: per our style guide, events and errors come after storage variables (and events come before errors)

src/L2/FlashblockIndex.sol
xenoliss: I don't feel strongly but most of the contracts in the repo do use the constant version.

jackchuma (approved):

PR #199, opened 2026-03-03
use a proof threshold and allow ZK proofs after TEE nullification

roger-bai-coinbase:


src/multiproof/AggregateVerifier.sol
```diff
@@ -260,12 +278,17 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
+        uint256 intermediateBlockInterval,
+        uint256 proofThreshold
     ) {
+        // Block interval and intermediate block interval must be positive and divisible.
         if (blockInterval == 0 || intermediateBlockInterval == 0 || blockInterval % intermediateBlockInterval != 0) {
             revert InvalidBlockInterval(blockInterval, intermediateBlockInterval);
         }
 
+        // Proof threshold must be between 1 and 2.
+        if (proofThreshold < 1 || proofThreshold > 2) revert InvalidProofThreshold();
```
leopoldjoy: Nit: might it be syntactically easier to read to instead just do `proofThreshold != 1 && proofThreshold != 2`?

src/multiproof/AggregateVerifier.sol
```diff
@@ -144,6 +147,10 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
     /// @notice The timestamp of the game's expected resolution.
     Timestamp public expectedResolution;
 
+    /// @notice The number of proofs provided.
+    /// @dev Can be negative if a ZK proof is nullified.
+    int256 public proofCount;
```
leopoldjoy: Couldn't we make this `int8`?

src/multiproof/AggregateVerifier.sol
```diff
@@ -530,9 +573,24 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
             abi.encodePacked(intermediateRootToProve)
         );
 
-        // Set the game as challenged so that child games can't resolve.
-        status = GameStatus.CHALLENGER_WINS;
-        // Refund the bond. This can override a challenge.
+        if (proofType == ProofType.ZK) {
+            // Since a ZK proof vetoes a TEE proof, we make the proof count negative for ZK nullifications.
+            // This ensures that the game cannot resolve if a TEE proof is provided later.
+            proofCount = -1;
```
leopoldjoy: Should we use a constant for this negative amount?

src/multiproof/AggregateVerifier.sol
```diff
@@ -709,12 +767,18 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
     /// @notice Updates the expected resolution timestamp.
     function _updateExpectedResolution() internal {
-        bool hasTee = proofTypeToProver[ProofType.TEE] != address(0);
-        bool hasZk = proofTypeToProver[ProofType.ZK] != address(0);
+        uint64 delay;
 
-        if (!hasTee && !hasZk) revert NoProofProvided();
+        if (proofCount >= 2) {
+            delay = FAST_FINALIZATION_DELAY;
+        } else if (proofCount > 0) {
```
leopoldjoy: Nit: syntactically it feels like the two conditionals should follow the same comparison practice (e.g. switch here to `... >= 1`).

src/multiproof/AggregateVerifier.sol
```diff
@@ -530,9 +573,24 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
+        if (proofType == ProofType.ZK) {
+            // Since a ZK proof vetoes a TEE proof, we make the proof count negative for ZK nullifications.
+            // This ensures that the game cannot resolve if a TEE proof is provided later.
+            proofCount = -1;
+
+            // Set the game as challenged so that child games can't resolve.
+            status = GameStatus.CHALLENGER_WINS;
+        } else if (proofType == ProofType.TEE) {
+            // The status is not updated here to still allow a ZK proof to be provided later.
+            proofCount -= 1;
```
leopoldjoy, line 580: Also here, should we use a constant for the deduction amount?

src/multiproof/AggregateVerifier.sol
```diff
@@ -481,6 +512,12 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
         // Set the game as challenged.
         status = GameStatus.CHALLENGER_WINS;
 
+        // Set the proof count to -1.
```
leopoldjoy: Nit: let's make this comment more instructive. Essentially we do this so that even if a TEE proof were to come in, in support of this proposal, it would be unable to reach any type of finality.

src/multiproof/AggregateVerifier.sol
```diff
@@ -509,6 +546,12 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
         if (proposedIntermediateRoot == intermediateRootToProve) revert IntermediateRootSameAsProposed();
 
         ProofType proofType = ProofType(uint8(proofBytes[0]));
+
+        // If TEE and ZK proofs are provided, can only nullify with ZK proof.
```
leopoldjoy: Do we still require this check now that we account via `proofCount`?

roger-bai-coinbase: Yes. Suppose we have a TEE + ZK proof, so finalization is in 1 day. Then if the TEE was nullified, the resolution should go back to 7 days, but currently doesn't.

We either keep this check or add logic that if we go from 2 proofs to 1, resolution increases by 6 days.

src/multiproof/AggregateVerifier.sol
roger-bai-coinbase: I don't think so. One proof is nullified so we just decrease the proof count by 1

src/multiproof/AggregateVerifier.sol
leopoldjoy: Wouldn't it be better to go with the latter in terms of recovery from TEE nullification, rather than just disable the whole game? What am I missing here?

PR #200, opened 2026-03-04
feat(CHAIN-3336): add Rust bindings and reorganize bindings directory

xenoliss:
## Summary

Reorganize contract bindings under a single `bindings/` directory with language-specific subdirectories:

- `bindings/go/` — existing Go bindings (BalanceTracker, FeeDisburser) moved from `bindings/`
- `bindings/rust/` — new Rust bindings crate (`base-contracts-bindings`) using Alloy's `sol!` macro, moved from `crates/bindings/`

The Rust crate exposes typed bindings (ABI + bytecode) generated from committed Forge artifacts, so downstream Rust consumers can depend on it via Cargo without needing Foundry.

### DIFF
--- a/snapshots/abi/NitroEnclaveVerifier.json
+++ b/snapshots/abi/NitroEnclaveVerifier.json
@@ -1,0 +1,1263 @@
+[
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "_owner",
+        "type": "address"
+      },
+      {
+        "internalType": "uint64",
+        "name": "_maxTimeDiff",
+        "type": "uint64"
+      },
+      {
+        "internalType": "bytes32[]",
+        "name": "_initializeTrustedCerts",
+        "type": "bytes32[]"
+      }
+    ],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "enum ZkCoProcessorType",
+        "name": "_zkCoProcessor",
+        "type": "uint8"
+      },
+      {
+        "internalType": "bytes4",
+        "name": "_selector",
+        "type": "bytes4"
+      },
+      {
+        "internalType": "address",
+        "name": "_verifier",
+        "type": "address"
+      }
+    ],
+    "name": "addVerifyRoute",
+    "outputs": [],
+    "stateMutability": "nonpayable",
+    "type": "function"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "bytes",
+        "name": "output",
+        "type": "bytes"
+      },
+      {
+        "internalType": "enum ZkCoProcessorType",
+        "name": "zkCoprocessor",
+        "type": "uint8"
+      },
+      {
+        "internalType": "bytes",
+        "name": "proofBytes",
+        "type": "bytes"
+      }
+    ],
+    "name": "batchVerify",
+    "outputs": [
+      {
+        "components": [
+          {
+            "internalType": "enum VerificationResult",
+            "name": "result",
+            "type": "uint8"
+          },
+          {
+            "internalType": "uint8",
+            "name": "trustedCertsPrefixLen",
+            "type": "uint8"
+          },
+          {
+            "internalType": "uint64",
+            "name": "timestamp",
+            "type": "uint64"
+          },
+          {
+            "internalType": "bytes32[]",
+            "name": "certs",
+            "type": "bytes32[]"
+          },
+          {
+            "internalType": "bytes",
+            "name": "userData",
+            "type": "bytes"
+          },
+          {
+            "internalType": "bytes",
+            "name": "nonce",
+            "type": "bytes"
+          },
+          {
+            "internalType": "bytes",
+            "name": "publicKey",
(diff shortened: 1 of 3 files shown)


bindings/go/balance_tracker.go
jackchuma, line 1: I'd be ok with just removing the go bindings altogether but am not too opinionated on it

xenoliss: I was not sure if they were already being used other external projects but I'm ok as well to remove them if not

bindings/rust/artifacts/FlashblockIndex.json
jackchuma, line 1: The `sol!` macro only needs the `abi` and `bytecode` fields. This artifact is 52KB but could be reduced to ~2KB by stripping unnecessary fields. This will compound as more contracts are added.

Suggestion: Strip the artifact to only `abi`, `bytecode.object`, and `deployedBytecode.object`, or add a script that generates minimal artifacts. The Alloy sol! macro with the JSON feature only reads abi and bytecode/deployedBytecode — everything else is dead weight in the repo.

bindings/rust/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "base-contracts-bindings"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+alloy-contract = "1.5.0"
```
jackchuma, line 7: `alloy-contract` doesn't appear to be used anywhere. Is this dep needed?

jackchuma (commented):
There's no CI check or script to verify that committed artifacts in `bindings/rust/artifacts/` match the current compiled output from `forge build`. If someone modifies `FlashblockIndex.sol` without regenerating the artifact, the Rust bindings will silently be out of sync.

Can we also add a CI step or a `just` recipe that:

1. Runs `forge build`
2. Compares `forge-artifacts/FlashblockIndex.sol/FlashblockIndex.json` with `bindings/rust/artifacts/FlashblockIndex.json`
3. Fails if they differ

This would be analogous to the existing `snapshots-check` pattern

bindings/rust/Cargo.toml
xenoliss: `alloy-contract` is required by the `#[sol(rpc)]` attribute

bindings/rust/artifacts/FlashblockIndex.json
xenoliss: Good catch, implemented a just recipe that does all that

justfile
```diff
@@ -222,6 +222,72 @@ snapshots-no-build: snapshots-abi-storage-no-build semver-lock-no-build
...
+  sol_path="{{SOL_PATH}}"
+
+  # Extract contract name (e.g. FlashblockIndex from src/L2/FlashblockIndex.sol)
+  contract=$(basename "$sol_path" .sol)
+
+  # Extract module (e.g. l2 from src/L2/FlashblockIndex.sol)
+  module=$(echo "$sol_path" | sed 's|^src/||' | xargs dirname | tr '[:upper:]' '[:lower:]')
+
+  # Convert ContractName to snake_case (e.g. FlashblockIndex -> flashblock_index)
+  snake=$(echo "$contract" | sed 's/\([A-Z]\)/_\L\1/g' | sed 's/^_//')
```
jackchuma: This is fragile

The sed-based conversion `sed 's/\([A-Z]\)/_\L\1/g' | sed 's/^_//'` doesn't handle consecutive uppercase sequences correctly. For example:
  - `ERC20` → `_e_r_c20` (incorrect, should be `erc20`)
  - `L1Block` → `_l1_block` → `l1_block` (happens to work by accident for leading uppercase)
  - `OPContractsManager` → `_o_p_contracts_manager` (incorrect)

This will produce wrong module names for contracts with acronyms, which are common in this codebase (e.g., `SuperchainERC20`, `OPContractsManager`). Consider a more robust approach or documenting this limitation.

xenoliss: Good catch thanks. Switched to a two-pass sed that handles acronym boundaries correctly. Successfully tested against all 170+ contracts in `src/,` all convert correctly.

bindings/rust/Cargo.lock
jackchuma, line 1: Is this file 100% needed? I wonder if CI still works without it. If so then it should be removed since it won't affect consumers of this library anyways

justfile
```diff
@@ -222,6 +222,72 @@ snapshots-no-build: snapshots-abi-storage-no-build semver-lock-no-build
...
+
+  # 2. Create module directory if needed
+  mkdir -p "$rust_dir"
+
+  # 3. Create .rs file
+  cat > "$rust_file" << EOF
+  use alloy_sol_types::sol;
+
+  sol!(
+      #[sol(rpc, abi)]
```
jackchuma, line 264: `abi` is not a documented attribute for the alloy `sol!` macro. Does including it here definitely change behavior?

xenoliss: it generates a dynamic ABI module (`alloy_json_abi`) for runtime ABI introspection. See [here](https://docs.rs/alloy-sol-macro/latest/alloy_sol_macro/macro.sol.html#attributes).

PR #201, opened 2026-03-05
Simplify and modularize aggregate verifier + verifiers

roger-bai-coinbase:
- TEE proof can be challenged with a ZK proof using intermediary roots
- Game can still progress after a TEE nullification or ZK nullification by providing a proof of the un-nulllified system
- Game can not progress if both proof systems nullified
- Bond for games that cannot progress are unlocked after 2 weeks from initialization time
- Nullifying strictly increases resolution time to allow time to intervene
- Challenges strictly increase resolution time to allow time to nullify the challenge
- Nullification automatically disables the nullified proof system, preventing further proof verifications from any game

### DIFF
--- a/src/multiproof/AggregateVerifier.sol
+++ b/src/multiproof/AggregateVerifier.sol
@@ -489,11 +489,11 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
         // This also prevents another challenge while the current challenge is in progress.
         if (proofTypeToProver[ProofType.ZK] != address(0)) revert AlreadyProven(ProofType.ZK);
 
-        _checkIntermediateRoot(intermediateRootIndex, intermediateRootToProve);
-
         // Can only challenge with a ZK proof.
         ProofType proofType = ProofType(uint8(proofBytes[0]));
         if (proofType != ProofType.ZK) revert InvalidProofType();
+
+        _checkIntermediateRoot(intermediateRootIndex, intermediateRootToProve);
 
         (bytes32 startingRoot, uint256 startingL2SequenceNumber, uint256 endingL2SequenceNumber) =
             _getStartingIntermediateRootAndL2SequenceNumbers(intermediateRootIndex);
@@ -779,7 +779,12 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
     /// @dev Should only occur if challenged or nullified.
     function _proofRefutedUpdate(ProofType proofType) internal {
         delete proofTypeToProver[proofType];
-        proofCount -= 1;
+
+        // Should not be possible, but just in case.
+        if (proofCount == 0) revert NotEnoughProofs();
+        unchecked {
+            proofCount -= 1;
+        }
 
         _increaseExpectedResolution();
     }
@@ -974,7 +979,8 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
     }
 
     /// @notice Checks if the intermediate root index is valid and that the intermediate root differs from the proposed
-    /// intermediate root. @param intermediateRootIndex The index of the intermediate root to check.
+    /// intermediate root.
+    ///@param intermediateRootIndex The index of the intermediate root to check.
     /// @param intermediateRootToProve The intermediate root that the proof claims to be correct.
     function _checkIntermediateRoot(uint256 intermediateRootIndex, bytes32 intermediateRootToProve) internal view {
         if (intermediateRootIndex >= intermediateOutputRootsCount()) revert InvalidIntermediateRootIndex();
--- a/src/multiproof/Verifier.sol
+++ b/src/multiproof/Verifier.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
 import { IAnchorStateRegistry } from "interfaces/dispute/IAnchorStateRegistry.sol";
--- a/snapshots/semver-lock.json
+++ b/snapshots/semver-lock.json
@@ -240,12 +240,12 @@
     "sourceCodeHash": "0x955bd0c9b47e43219865e4e92abf28d916c96de20cbdf2f94c8ab14d02083759"
   },
   "src/multiproof/AggregateVerifier.sol:AggregateVerifier": {
-    "initCodeHash": "0x24c7c851cb37c11121fcbfc20b2474d9f15122c020ac5b9894e371cf2f669c5b",
-    "sourceCodeHash": "0x4fb3fe897e702eb2d47cfaf67fa5b9651112580ed47028a5f5b3b810866635af"
+    "initCodeHash": "0xc3866b1d4515c9d7b0ac6679b182d836f79371402d9e649e301b24cf8ae8fade",
+    "sourceCodeHash": "0x3a079ea52a26c8c38fb0cb3e9a9ff6ec9648cf83786b65c0fc1161e949b8f7e0"
   },
   "src/multiproof/AggregateVerifier.sol:AggregateVerifier:dispute": {
-    "initCodeHash": "0x317136b944b743a70c447d78e362edc3e17cda44526ee97d90e66584ba64b21c",
-    "sourceCodeHash": "0x4fb3fe897e702eb2d47cfaf67fa5b9651112580ed47028a5f5b3b810866635af"
+    "initCodeHash": "0xe28eaeecda21594f6db23bb70127daa2b7b71debe38ce65b598f28d78d2561eb",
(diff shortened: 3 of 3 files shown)


src/multiproof/Verifier.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// @notice Verifies a proof.
+    /// @param proofBytes The proof.
+    /// @param imageId The image ID.
+    /// @param journal The journal.
+    /// @return valid Whether the proof is valid.
+    /// @dev This function must be overridden by the concrete verifier implementation.
+    ///      Ensure the modifier stays.
+    function verify(
```
jackchuma: We should be able to remove this since it's an abstract contract right? I don't think keeping this here with a comment saying to ensure the modifier stays is a best practice

src/multiproof/Verifier.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: UNLICENSED
```
leopoldjoy: Nit: I think we should likely keep the same licensing for all contracts:

```suggestion
// SPDX-License-Identifier: MIT
```

src/multiproof/AggregateVerifier.sol
```diff
@@ -983,6 +973,35 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
         }
     }
 
+    /// @notice Checks if the intermediate root index is valid and that the intermediate root differs from the proposed
+    /// intermediate root. @param intermediateRootIndex The index of the intermediate root to check.
```
leopoldjoy: Nit:
```suggestion
    /// intermediate root.
    /// @param intermediateRootIndex The index of the intermediate root to check.
```

src/multiproof/AggregateVerifier.sol
```diff
@@ -489,41 +484,48 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
-        (,, IDisputeGame game) = DISPUTE_GAME_FACTORY.gameAtIndex(gameIndex);
+        _checkIntermediateRoot(intermediateRootIndex, intermediateRootToProve);
 
-        // The game must be a valid game used to challenge.
-        if (!_isValidChallengingGame(game)) revert InvalidGame();
-
-        AggregateVerifier challengingGame = AggregateVerifier(address(game));
+        // Can only challenge with a ZK proof.
+        ProofType proofType = ProofType(uint8(proofBytes[0]));
+        if (proofType != ProofType.ZK) revert InvalidProofType();
```
leopoldjoy, line 494: Nit: in the case where this conditional evaluates to true, it would save gas to place it above the `_checkIntermediateRoot` check on line 492.

src/multiproof/Verifier.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    /// @notice Nullifies the verifier to prevent further proof verification.
+    /// @dev Should only occur if a soundness issue is found.
+    /// @dev Should only be callable by a proper dispute game.
+    function nullify() external override {
+        if (
+            !ANCHOR_STATE_REGISTRY.isGameProper(IDisputeGame(msg.sender))
+                || !ANCHOR_STATE_REGISTRY.isGameRespected(IDisputeGame(msg.sender))
+        ) revert NotProperGame();
+        nullified = true;
```
leopoldjoy, line 40: Just to confirm, since a single game's nullification permanently disables the verifier for all games (current and future) and no reversal mechanism exists, if a TEE verifier is nullified due to one bad game, for example, no new TEE proofs can ever be verified by any game until a new verifier is deployed and reconfigured, correct?

This seems correct to me, but want to categorically confirm.

src/multiproof/AggregateVerifier.sol
```diff
@@ -405,9 +391,10 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
             intermediateOutputRoots()
         );
 
-        _updateProvingData(proofType, gameCreator());
+        _proofVerifiedUpdate(proofType, gameCreator());
 
-        emit Proved(gameCreator(), proofType);
+        // Set the bond recipient to the creator. It can change if challenged successfully.
+        bondRecipient = gameCreator();
```
leopoldjoy, line 397: So if a challenger counters successfully but the game never resolves, the challenger doesn't get the bond, right? This behaviour seems correct to me, but wanted to confirm expectations.

src/multiproof/AggregateVerifier.sol
```diff
@@ -489,41 +484,48 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
-        proofCount = NEGATIVE_PROOF_COUNT;
+        // This is only in case the ZK proof is nullified, which would lower the proof count.
+        // If the ZK is nullified, we allow the remaining TEE proof to resolve.
+        // The expected resolution time can no longer be increased as both proof types have been submitted.
+        proofCount += 1;
 
-        // Update the expected resolution.
-        _updateExpectedResolution();
+        // We purposely increase the resolution to allow for a ZK nullification.
+        expectedResolution = Timestamp.wrap(uint64(block.timestamp + SLOW_FINALIZATION_DELAY));
```
leopoldjoy, line 522: To confirm, this enables TEE-only resolution following ZK nullification by reverting to the `SLOW_FINALIZATION_DELAY` timeout (thus potentially causing a total finalisation delay of up to 2 weeks), correct?

Again, this aligns with my expectations of the system, but want to confirm.

src/multiproof/AggregateVerifier.sol
```diff
@@ -766,35 +748,59 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
...
         uint64 newResolution = uint64(block.timestamp) + delay;
         expectedResolution = Timestamp.wrap(uint64(FixedPointMathLib.min(newResolution, expectedResolution.raw())));
     }
 
-    function _updateProvingData(ProofType proofType, address prover) internal {
-        proofTypeToProver[proofType] = prover;
+    /// @dev Should only occur if challenged or nullified.
+    function _proofRefutedUpdate(ProofType proofType) internal {
+        delete proofTypeToProver[proofType];
+        proofCount -= 1;
```
leopoldjoy: Observation: `proofCount` can't underflow since the `MissingProof` check before nullification always prevents it. However, could it make sense to nonetheless explicitly guard against underflow in the function itself to improve robustness / resilience? Thoughts?

src/multiproof/Verifier.sol
roger-bai-coinbase: Yes, I think an upgrade/redeployment should be required if we want to reenable a verifier

src/multiproof/AggregateVerifier.sol
roger-bai-coinbase: How would the game never resolve after a successful challenge? By successful, I mean the challenge is not later nullified

src/multiproof/AggregateVerifier.sol
roger-bai-coinbase: Yes, but the theoretical finalization delay could be forever. E.g. Initialize with ZK -> nullify with ZK -> provide a TEE proof years later

src/multiproof/AggregateVerifier.sol
leopoldjoy: Yes, I mean if its later nullified. I suppose this is the same case as [here](https://github.com/base/contracts/pull/201#discussion_r2911741798), yes?

src/multiproof/AggregateVerifier.sol
leopoldjoy: Sure, but assuming there's no bug in the TEE prover, and the collective Multiproving system is progressing on single-path TEE-only finality, then a TEE proof would need to be provided in order to enable L2 -> L1 finality to progress at all, no? Obviously if we transitioned to the permissioned game then this also wouldn't be relevant.

As I see it, this theoretical indefinite finalisation delay only applies if L2 -> L1 finality has altogether stopped progressing.

src/multiproof/AggregateVerifier.sol
roger-bai-coinbase: Yes, if the ZK challenge was later nullified, the bond is refunded back to the proposer

src/multiproof/AggregateVerifier.sol
roger-bai-coinbase: If we assume the TEE is providing proofs regularly, then 3 weeks would be the max finalization delay:

- Initialize with a TEE proof
- Challenge with a ZK proof after almost 7 days
- Nullify the ZK proof after almost 7 days
- Resolve with the existing TEE proof after 7 days

After any nullification, the resolution is increased by 7 days to allow time to react, such as later nullifications for the other proof system, time to blacklist, etc.

leopoldjoy: Ok, I think this is acceptable behaviour.

PR #206, opened 2026-03-12
Add custom NitroEnclaveVerifier contract and update multiproof contracts

roger-bai-coinbase:
- Add NitroEnclaveVerifier with ZK-based AWS Nitro attestation verification
- Add tests for NitroEnclaveVerifier
- Add proof submitter configurability to NitroEnclaveVerifier
- Update TEEVerifier, MockVerifier, and deploy scripts accordingly
- Update snapshots and semver-lock

Made-with: Cursor

### DIFF
--- a/scripts/deploy/DeployImplementations.s.sol
+++ b/scripts/deploy/DeployImplementations.s.sol
@@ -44,9 +44,7 @@ import { DeployUtils } from "scripts/libraries/DeployUtils.sol";
 import { Solarray } from "scripts/libraries/Solarray.sol";
 import { ChainAssertions } from "scripts/deploy/ChainAssertions.sol";
 import { DevFeatures } from "src/libraries/DevFeatures.sol";
-import {
-    INitroEnclaveVerifier
-} from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
+import { INitroEnclaveVerifier } from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
 import { TEEProverRegistry } from "src/multiproof/tee/TEEProverRegistry.sol";
 import { MockVerifier } from "src/multiproof/mocks/MockVerifier.sol";
 import { TEEVerifier } from "src/multiproof/tee/TEEVerifier.sol";
--- a/scripts/multiproof/DeployDevNoNitro.s.sol
+++ b/scripts/multiproof/DeployDevNoNitro.s.sol
@@ -52,9 +52,7 @@ pragma solidity 0.8.15;
  * ══════════════════════════════════════════════════════════════════════════════════
  */
 
-import {
-    INitroEnclaveVerifier
-} from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
+import { INitroEnclaveVerifier } from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
 import { TransparentUpgradeableProxy } from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 import { Script } from "forge-std/Script.sol";
 import { console2 as console } from "forge-std/console2.sol";
--- a/scripts/multiproof/DeployDevWithNitro.s.sol
+++ b/scripts/multiproof/DeployDevWithNitro.s.sol
@@ -86,9 +86,7 @@ pragma solidity 0.8.15;
  * ══════════════════════════════════════════════════════════════════════════════════
  */
 
-import {
-    INitroEnclaveVerifier
-} from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
+import { INitroEnclaveVerifier } from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
 import { TransparentUpgradeableProxy } from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 import { Script } from "forge-std/Script.sol";
 import { console2 as console } from "forge-std/console2.sol";
--- a/src/multiproof/mocks/MockDevTEEProverRegistry.sol
+++ b/src/multiproof/mocks/MockDevTEEProverRegistry.sol
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
-import {
-    INitroEnclaveVerifier
-} from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
+import { INitroEnclaveVerifier } from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
 import { EnumerableSetLib } from "@solady-v0.0.245/utils/EnumerableSetLib.sol";
 
 import { TEEProverRegistry } from "src/multiproof/tee/TEEProverRegistry.sol";
--- a/test/multiproof/TEEProverRegistry.t.sol
+++ b/test/multiproof/TEEProverRegistry.t.sol
@@ -6,9 +6,7 @@ import { Test } from "forge-std/Test.sol";
 import { TransparentUpgradeableProxy } from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 import { ProxyAdmin } from "src/universal/ProxyAdmin.sol";
 
-import {
-    INitroEnclaveVerifier
-} from "interfaces/multiproof/tee/INitroEnclaveVerifier.sol";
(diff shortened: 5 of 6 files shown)


src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
...
+
+        _zkVerifierRoutes[_zkCoProcessor][_selector] = FROZEN;
+        emit ZkRouteWasFrozen(_zkCoProcessor, _selector);
+    }
+
+    /**
+     * @dev Sets the proof submitter address
+     * @param _proofSubmitter The address of the proof submitter
+     */
+    function setProofSubmitter(address _proofSubmitter) external onlyOwner {
```
leopoldjoy, line 518: `setProofSubmitter` allows `address(0)`, which would effectively brick `verify()` and `batchVerify()` — no one can have msg.sender == address(0). Consider adding: `require(_proofSubmitter != address(0), "Proof submitter cannot be zero address");`

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
...
+    address public proofSubmitter;
+
+    /// @dev Configuration mapping for each supported ZK coprocessor type
+    mapping(ZkCoProcessorType => ZkCoProcessorConfig) public zkConfig;
+
+    /// @dev Mapping of trusted intermediate certificate hashes (excludes root certificate)
+    mapping(bytes32 trustedCertHash => bool) public trustedIntermediateCerts;
+
+    /// @dev Maximum allowed time difference in seconds for attestation timestamp validation
+    uint64 public maxTimeDiff;
```
leopoldjoy, line 65: Not necessarily actionable, but currently there's no way to update `maxTimeDiff` after construction. So if operational requirements change (e.g., need to tighten or relax the time window), the contract must be fully redeployed. I suppose this is fine, but wanted to highlight this decision.

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
...
+     */
+    function verify(
+        bytes calldata output,
+        ZkCoProcessorType zkCoprocessor,
+        bytes calldata proofBytes
+    )
+        external
+        returns (VerifierJournal memory journal)
+    {
+        require(msg.sender == proofSubmitter, "Only the proof submitter can verify proofs");
```
leopoldjoy: I noticed that the contract mixes string reverts (`require(msg.sender == proofSubmitter, "Only the proof submitter can verify proofs")`, `revert("Not implemented")`, `revert("Certificate not found...")`) with custom errors (`revert Unknown_Zk_Coprocessor()`, `revert ZkRouteFrozen(...)`). String reverts cost significantly more gas. Since the interface already defines custom errors, consider adding new custom errors for the custom logic (e.g., `error Unauthorized()`, etc.) for consistency and gas savings.

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
...
+        ZkCoProcessorType,
+        bytes32,
+        bytes32,
+        bytes calldata
+    )
+        external
+        pure
+        returns (VerifierJournal[] memory)
+    {
+        revert("Not implemented");
```
leopoldjoy: Why must we follow the interface if we don't need this functionality? If we do need to follow it, then we should add `@notice NatSpec` documenting that these are intentionally disabled and why.

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
+//SPDX-License-Identifier: Apache2.0
```
leopoldjoy: Do we need to stick with Apache-2.0 here?

test/multiproof/NitroEnclaveVerifier.t.sol
leopoldjoy, line 1: The following revert paths have no test coverage:
- `Unknown_Zk_Coprocessor()` error (invalid coprocessor type)
- `ZkVerifierNotConfigured` error (no verifier set, no route match)
- `ZkRouteFrozen` during an actual verify() call (vs. just the `getZkVerifier` view function)
- Owner-only checks on `updateVerifierId`, `updateAggregatorId`, `removeVerifierId`, `removeAggregatorId`, `addVerifyRoute`, `freezeVerifyRoute`

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,660 @@
...
+        internal
+        view
+    {
+        // Resolve the verifier address (route-specific or default)
+        address verifier = _resolveZkVerifier(zkCoprocessor, proofBytes);
+
+        if (zkCoprocessor == ZkCoProcessorType.RiscZero) {
+            IRiscZeroVerifier(verifier).verify(proofBytes, programId, sha256(output));
+        } else if (zkCoprocessor == ZkCoProcessorType.Succinct) {
+            ISP1Verifier(verifier).verifyProof(programId, output, proofBytes);
```
leopoldjoy, line 692: We want the system to support both RiscZero and SP1 out of the box, correct? If so, then we should also add code path coverage for SP1 into the tests file below as it currently has zero test coverage.

test/multiproof/NitroEnclaveVerifier.t.sol
```diff
@@ -0,0 +1,614 @@
...
+    bytes32 public constant ROOT_CERT = keccak256("root-cert");
+    bytes32 public constant INTERMEDIATE_CERT_1 = keccak256("intermediate-cert-1");
+    bytes32 public constant INTERMEDIATE_CERT_2 = keccak256("intermediate-cert-2");
+    bytes32 public constant VERIFIER_ID = keccak256("verifier-id");
+    bytes32 public constant AGGREGATOR_ID = keccak256("aggregator-id");
+    bytes32 public constant VERIFIER_PROOF_ID = keccak256("verifier-proof-id");
+
+    uint64 public constant MAX_TIME_DIFF = 3600; // 1 hour
+
+    function setUp() public {
```
leopoldjoy, line 37: Nit: Tests run at default Foundry `block.timestamp` (1), which is unrealistic for timestamp validation logic. Consider adding `vm.warp(1_700_000_000)` or similar in `setUp()` to test with realistic timestamps.

src/multiproof/tee/NitroEnclaveVerifier.sol
roger-bai-coinbase: Would it be better to make our own interface then? Unsure how we want to handle this

src/multiproof/tee/NitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,732 @@
+// SPDX-License-Identifier: Apache2.0
+pragma solidity ^0.8.0;
+
+import { Ownable } from "@solady/auth/Ownable.sol";
+import { EnumerableSet } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import {
+    INitroEnclaveVerifier,
```
leopoldjoy, line 7: [`TEEProverRegistry`](https://github.com/base/contracts/blob/main/src/multiproof/tee/TEEProverRegistry.sol#L11) and [`MockDevTEEProverRegistry`](https://github.com/base/contracts/blob/main/src/multiproof/mocks/MockDevTEEProverRegistry.sol#L6) still import `INitroEnclaveVerifier` / `VerifierJournal` / `ZkCoProcessorType` from the upstream lib, while this contract implements the custom interface. These are distinct Solidity types — we should probably update those contracts to use the custom interface.

interfaces/multiproof/tee/INitroEnclaveVerifier.sol
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * @dev Sets the proof submitter address
+     * @param _proofSubmitter The address of the proof submitter
+     *
+     * Requirements:
+     * - Only callable by contract owner
+     * - Address must not be zero
+     */
+    function setProofSubmitter(address _proofSubmitter) external;
```
leopoldjoy, line 263: Nit: `setProofSubmitter` is in the interface but the `proofSubmitter()` getter is not, so consumers using the interface type can set it but not read it. Let's also add `function proofSubmitter() external view returns (address);` to the query section?

leopoldjoy (approved):

PR #207, opened 2026-03-17
feat(CHAIN-3336): add Rust bindings and reorganize bindings directory

xenoliss:
Replaces https://github.com/base/contracts/pull/200

## Summary

Set up Rust contract bindings under `bindings/rust/` as a standalone crate and remove the legacy Go bindings.

- Remove the old Go bindings under `bindings/go/`.
- Add the Rust bindings crate (`base-contracts-bindings`) under `bindings/rust/`.
- Generate Rust bindings for all contracts defined under `src/` using `forge bind`.
- Use a single global regex matcher derived from `src/**/*.sol` filenames to restrict `forge bind` to source contracts only.

The Rust crate is set up so downstream Rust consumers can depend on it directly via Cargo without needing Foundry locally.

This PR also adds a `just bindings-generate` workflow that regenerates the committed Rust bindings from the current `src/` contract set.

To keep the new Rust bindings reliable, this PR adds validation recipes and CI coverage:

- `just bindings-check-no-build` verifies that committed Rust bindings stay in sync with `forge-artifacts` by checking `forge bind` output without rewriting files.
- `just bindings-check` runs the same consistency check and then `cargo check` in `bindings/rust` to ensure the crate compiles.
- CI now installs Rust and runs `just bindings-check`.


jackchuma (dismissed):
Foundry actually provides a utility for this. This PR can basically be reduced to

```
  bindings-generate: build-source
    forge bind --select "..." --crate-name base-contracts-bindings -b bindings/rust --overwrite

  bindings-check:
    forge bind --select "..." --crate-name base-contracts-bindings -b bindings/rust --skip-build
```

https://www.getfoundry.sh/reference/forge/bind#forge-bind

PR #276, opened 2026-05-11
feat(NitroEnclaveVerifier): durable on-chain certificate revocation [non-urgent, post-Multiproving]

leopoldjoy:
## Priority

**Not urgent.** This change can ship **after** the Multiproving launch (Base V2 upgrade). Multiproving is the primary mitigation for any single-verifier gap; this PR closes a defense-in-depth hole that would otherwise require either a CA key compromise (Tier B) or an honest-but-racing registrar/admin scenario (Tier A) to exploit. There is no known active exploitation path against the current production configuration.

## Summary

- Adds a persistent `revokedCerts` mapping that survives the `_cacheNewCert` suffix rewrite, so a previously revoked intermediate cert path cannot be silently re-trusted by submitting a new attestation that re-derives the same accumulated path digest.
- Adds `unrevokeCert(bytes32) onlyOwner` for governed re-trust (two-step: unrevoke, then re-cache via the next successful `verify`).
- Reverts in `_cacheNewCert` (Pass 1) and `_verifyJournal` (Pass 2) when a sentinel-keyed digest is encountered, plus an early short-circuit in `checkTrustedIntermediateCerts`.
- Bumps semver `0.3.0` → `0.4.0` (additive — no storage layout changes that move existing slots; only a new mapping appended).

## Background — Immunefi #75608 / CHAIN-4194

`_cacheNewCert` previously rewrote suffix entries unconditionally. If the prefix length is `1` (the production registrar default), revoking an intermediate via `revokeCert` only zeroed the cached entry — a subsequent attestation re-deriving the same accumulated path digest would re-cache and re-trust it. The new `revokedCerts[digest] = true` sentinel is checked on every code path that would otherwise mark a cert as trusted, making revocation durable across attestations.

Triage: **Low.** Tier B requires CA key compromise to forge a new chain that hashes back to the revoked digest; Tier A requires a registrar honest-mistake / admin race. Both are mitigated in production today by operational controls and (post-launch) by Multiproving. This PR is hardening, not an incident response.

## Changes

- `src/L1/proofs/tee/NitroEnclaveVerifier.sol`
  - New storage: `mapping(bytes32 => bool) public revokedCerts` (appended, layout-safe).
  - `revokeCert`: sets sentinel + zeroes any existing cached entry, emits `CertRevoked`.
  - `unrevokeCert`: clears sentinel only, emits `CertUnrevoked` (re-trust requires a fresh `verify` call to re-cache).
  - `_cacheNewCert`: skips writing trusted suffix entries whose accumulated digest is sentineled.
  - `_verifyJournal`: rejects with `CertificateRevoked(digest)` when traversal encounters a sentineled digest.
  - `checkTrustedIntermediateCerts`: short-circuits to `false` if any digest in the chain is sentineled.
- `interfaces/L1/proofs/tee/INitroEnclaveVerifier.sol`: declares `revokedCerts`, `unrevokeCert`, `CertUnrevoked` event, `CertificateNotRevoked` error.
- `test/L1/proofs/NitroEnclaveVerifier.t.sol`: 8 new tests covering revoke→re-attest rejection, unrevoke flow, sentinel collision behaviour, owner gating, and event emission. All 81/81 NitroEnclaveVerifier and 366/366 L1 proofs tests pass.
- Snapshots regenerated for `NitroEnclaveVerifier` only.

## Compatibility

- **Storage layout:** strictly append-only. Existing slots untouched. Safe for in-place upgrade.
- **Off-chain registrar:** the companion offchain PR is **backwards compatible** with the current (unpatched) deployed contract — it fail-opens (warns + counter-bumps) when `revokedCerts` is not present, so the offchain change can be deployed in either order relative to this onchain upgrade.

## Out of Scope / Notes

- Snapshot regen also produced unrelated drift in `snapshots/abi/AggregateVerifier.json`; intentionally **not** included in this PR.

### DIFF
--- a/src/L1/proofs/tee/NitroEnclaveVerifier.sol
+++ b/src/L1/proofs/tee/NitroEnclaveVerifier.sol
@@ -70,6 +70,19 @@ contract NitroEnclaveVerifier is Ownable, INitroEnclaveVerifier, ISemver {
     /// @dev Mapping from ZkCoProcessorType to its corresponding verifierProofId representation
     mapping(ZkCoProcessorType => bytes32) private _verifierProofIds;
 
+    /// @dev Persistent revocation sentinel for intermediate certificates.
+    ///
+    /// `revokeCert` zeroes `trustedIntermediateCerts[certHash]`, but the suffix-cache
+    /// path in `_cacheNewCert` would otherwise restore that entry on the next
+    /// successful verification whose chain traverses the revoked hash. This
+    /// mapping survives `_cacheNewCert` overwrites and is consulted in
+    /// `_verifyJournal`, `_cacheNewCert`, and `checkTrustedIntermediateCerts`,
+    /// making `revokeCert` durable independently of the journal's `trustedCertsPrefixLen`.
+    ///
+    /// Re-trust requires an explicit `unrevokeCert` admin call; it is never
+    /// granted as a side effect of verification.
+    mapping(bytes32 => bool) public revokedCerts;
+
     // ============ Custom Errors ============
 
     /// @dev Error thrown when an unsupported or unknown ZK coprocessor type is used
@@ -114,6 +127,9 @@ contract NitroEnclaveVerifier is Ownable, INitroEnclaveVerifier, ISemver {
     /// @dev Thrown when caller is neither the owner nor the revoker
     error CallerNotOwnerOrRevoker();
 
+    /// @dev Thrown when `unrevokeCert` is called for a hash that is not currently revoked
+    error CertificateNotRevoked(bytes32 certHash);
+
     // ============ Events ============
 
     /// @dev Emitted when a new verifier program ID is added/updated
@@ -145,6 +161,9 @@ contract NitroEnclaveVerifier is Ownable, INitroEnclaveVerifier, ISemver {
 
     /// @dev Event emitted when a certificate is revoked
     event CertRevoked(bytes32 certHash);
+
+    /// @dev Event emitted when a previously revoked certificate is explicitly re-trusted
+    event CertUnrevoked(bytes32 certHash);
 
     /// @dev Event emitted when the maximum time difference is updated
     event MaxTimeDiffUpdated(uint64 newMaxTimeDiff);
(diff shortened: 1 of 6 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

src/L1/proofs/tee/NitroEnclaveVerifier.sol
```diff
@@ -570,10 +624,17 @@ contract NitroEnclaveVerifier is Ownable, INitroEnclaveVerifier, ISemver {
...
+     *
+     * Revoked entries are skipped: once `revokedCerts[certHash]` is set by `revokeCert`,
+     * no successful verification will silently restore the cache, regardless of the
+     * journal's `trustedCertsPrefixLen`. Re-trust requires an explicit `unrevokeCert`.
      */
     function _cacheNewCert(VerifierJournal memory journal) internal {
         for (uint256 i = journal.trustedCertsPrefixLen; i < journal.certs.length; i++) {
             bytes32 certHash = journal.certs[i];
+            if (revokedCerts[certHash]) {
+                continue;
```
roger-bai-coinbase: Why continue here? Later certs depend on the revoked cert so I don't see why they should be trusted

leopoldjoy: Good catch — fixed in ba3c25cb. Switched `continue` → `break`. You're right that descendants of a revoked cert inherit trust from a revoked parent, so we shouldn't cache them either. Also aligns with `checkTrustedIntermediateCerts`, which already `break`s on the same condition.

Worth noting: in the current control flow this branch is unreachable because `_verifyJournal` Pass 2 already rejects any journal whose suffix contains a revoked digest before `_cacheNewCert` is invoked, so the change is defense-in-depth / correctness rather than a behavioral fix. Updated the doc comment to make that explicit.

roger-bai-coinbase (approved):

PR #328, opened 2026-06-04
feat: add proveAndFinalizeWithdrawalTransaction for L3 immediate TEE finality

robriks:
This PR demonstrates the minimalist approach for solidity-layer support of the privacy appchain as an L3 settling to the Base L2. The changes are as simple as possible, adjusting only the parameters necessary for the Base stack to be used for an L3 settling to L2, rather than L2 (Base) settling to L1 (Ethereum).

## Summary
- Adds `proveAndFinalizeWithdrawalTransaction` to `OptimismPortal2`, enabling atomic prove-and-finalize in a single transaction for L3 chains with TEE-backed immediate finality (where `PROOF_MATURITY_DELAY_SECONDS == 0`)
- Extracts shared withdrawal proof verification logic into `_verifyWithdrawalProof` internal helper, used by both `proveWithdrawalTransaction` and the new combined function
- Makes `SLOW_FINALIZATION_DELAY` and `FAST_FINALIZATION_DELAY` immutable constructor parameters in `AggregateVerifier` (via `FinalizationDelays` struct) to support L3 configurations with zero-delay finalization
- Adds `canProveAndFinalize` view function for off-chain tooling to check preconditions
- Adds `deploy-config/local-tee.json` for local TEE-enabled devnet deployments

## Testing
- [x] New tests for `proveAndFinalizeWithdrawalTransaction` happy path and all revert conditions
- [x] Existing `proveWithdrawalTransaction` and `finalizeWithdrawalTransaction` tests pass (refactor is behavior-preserving)
- [x] `AggregateVerifier` tests updated for immutable finalization delays
- [x] Deploy script and config tests updated


scripts/deploy/SystemDeploy.s.sol
```diff
@@ -1057,8 +1066,12 @@ contract SystemDeploy is Script {
...
     function _assertValidImplementationInput(ImplementationInput memory _input) internal pure {
         require(_input.withdrawalDelaySeconds != 0, "SystemDeploy: withdrawalDelaySeconds not set");
-        require(_input.proofMaturityDelaySeconds != 0, "SystemDeploy: proofMaturityDelaySeconds not set");
-        require(_input.disputeGameFinalityDelaySeconds != 0, "SystemDeploy: finality delay not set");
+        if (_input.proofMaturityDelaySeconds == 0) {
+            console.log("WARNING: proofMaturityDelaySeconds is 0 - immediate finality mode enabled");
+        }
+        if (_input.disputeGameFinalityDelaySeconds == 0) {
+            console.log("WARNING: disputeGameFinalityDelaySeconds is 0 - immediate finality mode enabled");
+        }
```
dguenther, line 1081: Maybe some sort of configuration guard would be useful to make sure deployers either set all the 'immediate finality mode' flags or none of them?

src/L1/OptimismPortal2.sol
```diff
@@ -551,6 +524,149 @@ contract OptimismPortal2 is Initializable, ResourceMetering, ReinitializableBase
...
+        bytes[] calldata _withdrawalProof
+    )
+        external
+    {
+        // Cannot prove/finalize withdrawal transactions while the system is paused.
+        _assertNotPaused();
+
+        // This function bypasses the proof maturity delay check enforced by checkWithdrawal.
+        // It is only safe to use on chains configured for immediate TEE-backed finality.
+        if (PROOF_MATURITY_DELAY_SECONDS != 0) {
```
dguenther, line 549: Does this need to check `anchorStateRegistry.disputeGameFinalityDelaySeconds() != 0`? Not sure if that's relevant here

dguenther (approved):

src/L1/proofs/AggregateVerifier.sol
```diff
@@ -270,7 +277,8 @@ contract AggregateVerifier is Clone, ReentrancyGuard, ISemver {
         bytes32 configHash,
         uint256 l2ChainId,
         uint256 blockInterval,
-        uint256 intermediateBlockInterval
+        uint256 intermediateBlockInterval,
+        FinalizationDelays memory delays
```
wbj-cb, line 281: I think its more clear to leave these as individual arguments `slowFinalizationDelay`, `fastFinalizationDelay` rather than a packed struct. Reading the constructor I was confused what the struct was

scripts/deploy/SystemDeploy.s.sol
wbj-cb, line 1081: If this is meant to be merged and deployable: can't use console.log

src/L1/proofs/AggregateVerifier.sol
robriks: I agree, but separating them into individual arguments results in a stack too deep error so I went with a struct. 

another option would be to pack the variables into a single slot using two `uint160` types, though that seems a bit more cumbersome than a dedicated struct to me

scripts/deploy/SystemDeploy.s.sol
robriks: I see why these can be seen as not particularly useful (threw them in for visibility there since they are replacing pre-existing checks for 0 which we now want to allow), but don't quite follow why they make it undeployable. how did you mean?

happy to change/remove these either way !

src/L1/OptimismPortal2.sol
robriks: it's relevant! admittedly would require a pretty borked up config at deployment time but definitely worth explicitly checking so that only valid states are set at config.

implemented in 02dbda38

scripts/deploy/SystemDeploy.s.sol
wbj-cb: they don't do anything on a contract deployed on a real node. iirc they rely on anvil/hardhat only opcodes to actually emit console logs

robriks: found a way to bring back the zero checks so we don't need warnings at all in f356f7dc !

robriks: implemented in f356f7dc

PR #332, opened 2026-06-09
chore: Make SystemDeploy salt mixer configurable

jowparks:
## Summary
- add a deploy config saltMixer value with DEPLOY_SALT_MIXER fallback
- use the configured salt mixer when SystemDeploy derives OP chain CREATE2 addresses

## Test plan
- forge build
- forge test --match-contract SystemDeploy_Test


robriks (approved):
solid !

PR #341, opened 2026-06-15
support optional dev environment multiproof contracts (TEE bypass) in…

robriks:
## Problem

The L3 devnet deploys the full OP Stack via `SystemDeploy` but with `multiproofConfigHash = 0`, skipping multiproof entirely. Without an `AggregateVerifier` registered on the `DisputeGameFactory`, dispute games can't be created and TEE instant-finality withdrawals (`proveAndFinalizeWithdrawalTransaction`) don't work. This is the primary gap between the current devnet and a functional L3 with TEE-backed withdrawals.

Enabling multiproof today requires real Nitro enclave and SP1 verifier contracts deployed on-chain, infrastructure we don't have wired in yet. 

## Solution

When `multiproofConfigHash` is non-zero (multiproof enabled) but `nitroEnclaveVerifier` is `address(0)` (no Nitro infra), `SystemDeploy` now deploys:

- `DevTEEProverRegistry` instead of `TEEProverRegistry`; allows `addDevSigner()` without attestation
- `MockVerifier` instead of `ZKVerifier`; auto-accepts any ZK proof (a follow-up PR will deprecate ZK path entirely)

Everything else is unchanged: real `DisputeGameFactory`, `AnchorStateRegistry`, `DelayedWETH`, and the TEEProverRegistry proxy sits behind the chain's `ProxyAdmin` for a clean upgrade to the production impl when Nitro is ready.

No new config fields — the mode is inferred from existing config values.

## Why not a separate deploy script?

- **`DeployDevNoNitro` standalone** — Deploys its own `DisputeGameFactory`/`MockAnchorStateRegistry` — disconnected from the real system
- **This PR** — Single atomic deploy, all real infra, proxy-upgradeable to production

## Testing

- [x] Existing `test_deploy_withoutManagerAddress_succeeds` (production path) still passes
- [x] New `test_deploy_devMultiproof_succeeds` — deploys with zero nitro/sp1, verifies `addDevSigner` works, confirms `AggregateVerifier` references real `DelayedWETH`, game type registered on factory
- [x] All 7 tests pass

### DIFF
--- a/scripts/deploy/SystemDeploy.s.sol
+++ b/scripts/deploy/SystemDeploy.s.sol
@@ -85,6 +85,7 @@ contract SystemDeploy is Script {
         ISP1Verifier sp1Verifier;
         address teeProposer;
         address teeChallenger;
+        address devTeeSigner;
         address guardian;
         address incidentResponder;
         uint64 slowFinalizationDelay;
@@ -274,6 +275,7 @@ contract SystemDeploy is Script {
             sp1Verifier: ISP1Verifier(cfg.sp1Verifier()),
             teeProposer: cfg.teeProposer(),
             teeChallenger: cfg.teeChallenger(),
+            devTeeSigner: cfg.devTeeSigner(),
             guardian: cfg.superchainConfigGuardian(),
             incidentResponder: cfg.superchainConfigIncidentResponder(),
             slowFinalizationDelay: cfg.slowFinalizationDelay(),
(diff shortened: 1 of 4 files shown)


scripts/deploy/SystemDeploy.s.sol
```diff
@@ -1084,16 +1098,14 @@ contract SystemDeploy is Script {
         return _input.multiproofConfigHash != bytes32(0);
     }
 
+    function _isDevMultiproof(ImplementationInput memory _input) internal pure returns (bool) {
```
hughy, line 1109: It doesn't look like this is called anywhere

wbj-cb:
Is there a way to make this explicitly controlled by a bool rather than conditionally on the input addresses?

scripts/deploy/SystemDeploy.s.sol
robriks: lol you're right, thanks claude. 

this is supposed to be the bool utility that defines the dev multiproof mode, fixed in e93f110a

wbj-cb (approved):

wbj-cb (approved):

PR #342, opened 2026-06-15
feat(PRIV-1977): deprecate zk proof path in dev multiproof mode

robriks:
## Summary

- Replaces `MockVerifier` deployment with `IVerifier(address(0xdead))` as the ZK verifier in dev multiproof mode (`sp1Verifier == address(0)`), making all ZK interactions hard-revert at the EVM level
- Adds 4 invariant tests proving the ZK path is permanently bricked: 
  - ZK proof submission reverts, 
  - ZK challenge reverts, 
  - ZK via `verifyProposalProof` reverts, and 
  - TEE-only resolution succeeds as `DEFENDER_WINS`

## Motivation

`MockVerifier` auto-accepts all ZK proofs, so anyone could submit a ZK challenge against a TEE-proved game. We've enabled instant TEE-trusted finality so the ZK path needs to be disabled entirely. 

Pointing `ZK_VERIFIER` at `address(0xdead)` (a codeless burn address) causes the Solidity ABI decoder to unconditionally revert on empty returndata, structurally disabling the ZK path without modifying `AggregateVerifier.sol`.

## Test plan

- [x] `forge test --match-path test/deploy/SystemDeploy.t.sol -vvv` — all 11 tests pass
- [x] `test_zkProofSubmission_reverts` — ZK proof at game creation reverts
- [x] `test_zkChallenge_reverts` — `challenge()` with ZK proof reverts  
- [x] `test_zkProofViaVerifyProposalProof_reverts` — `verifyProposalProof()` with ZK reverts
- [x] `test_resolve_succeedsWithTEEOnly` — TEE-only game resolves `DEFENDER_WINS`

### DIFF
--- a/scripts/deploy/SystemDeploy.s.sol
+++ b/scripts/deploy/SystemDeploy.s.sol
@@ -40,7 +40,6 @@ import { INitroEnclaveVerifier } from "interfaces/L1/proofs/tee/INitroEnclaveVer
 import { ISP1Verifier } from "interfaces/L1/proofs/zk/ISP1Verifier.sol";
 import { ZKVerifier } from "src/L1/proofs/zk/ZKVerifier.sol";
 import { DevTEEProverRegistry } from "test/mocks/MockDevTEEProverRegistry.sol";
-import { MockVerifier } from "test/mocks/MockVerifier.sol";
 import { Constants } from "src/libraries/Constants.sol";
 import { SemverComp } from "src/libraries/SemverComp.sol";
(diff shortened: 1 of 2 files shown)


dguenther (approved):

hughy (approved):
I think the only latent suggestion is that we add a guard (maybe around chainid?) so that dev configs can only be deployed on dev/test chains. Not a blocker since we are only doing dev chain POC right now!

PR #353, opened 2026-06-26
feat(L1): add ProtocolVersions upgrade schedule contract

PelleKrab:
## Summary

Adds `ProtocolVersions`, a Security Council-controlled L1 contract that
maintains an ordered registry of L2 upgrade activation timestamps and
commits to a reproducible `scheduleId` hash chain.

- **Registry**: upgrades are registered by name (e.g. `"canyon"`) with an
  owner-assigned protocol version; timestamps are set/cleared separately
  by the owner
- **chainTeam role**: a secondary address (appointed/revocable by the owner)
  that can only push already-scheduled activations further into the future
  via `delayTimestamp`, it cannot register, clear, or pull timestamps earlier
- **scheduleId**: a cumulative keccak256 chain over `(l2ChainId, address,
  ordered keys, current timestamps)`.
- **getSchedule**: single `eth_call` view returning the full ordered schedule
  (name, timestamp, protocol version, per-entry hash) for off-chain consumers

## Test plan

- [x] `forge test --match-path test/L1/ProtocolVersions.t.sol` passes
- [x] Confirm `scheduleId` reproducibility manually: compute chain from
  `(l2ChainId, address, ordered keys, timestamps)` and compare to
  `scheduleId()` return value
- [x] Verify `delayTimestamp` cannot be used to pull an activation earlier
  or to schedule a fresh one
- [x] Verify `setTimestamp(id, 0)` correctly clears a pending timestamp and
  reverts the `scheduleId` to the post-registration value

### DIFF
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -8,12 +8,6 @@ import { IReinitializableBase } from "interfaces/universal/IReinitializableBase.
 /// @title IProtocolVersions
 /// @notice Interface for the ProtocolVersions upgrade schedule contract.
 interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBase {
-    struct Upgrade {
-        uint256 id;
-        uint64 timestamp;
-        bytes32 scheduleId;
-    }
-
     error ProtocolVersions_UnknownUpgrade(uint256 id);
     error ProtocolVersions_InvalidProtocolVersion();
     error ProtocolVersions_ActivationAlreadyPassed(uint256 id, uint64 activationTimestamp);
@@ -26,7 +20,7 @@ interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBa
     event UpgradeRegistered(uint256 indexed id);
     event MinimumProtocolVersionUpdated(uint256 indexed protocolVersion);
     event TimestampSet(uint256 indexed id, uint256 timestamp);
-    event ScheduleIdUpdated(bytes32 indexed newScheduleId, uint256 indexed blockNumber);
+    event ScheduleIdUpdated(bytes32 indexed newScheduleId);
     event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
     event Initialized(uint8 version);
 
@@ -37,10 +31,10 @@ interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBa
     function scheduleId(uint256 id) external view returns (bytes32);
     function minimumProtocolVersion() external view returns (uint256);
 
-    function getSchedule() external view returns (Upgrade[] memory);
+    function getSchedule() external view returns (uint64[] memory);
 
     function initialize(address _chainTeam) external;
-    function registerUpgrade() external returns (uint256);
+    function registerUpgrade(uint64 timestamp, uint256 minProtocolVersion) external returns (uint256);
     function setMinimumProtocolVersion(uint256 protocolVersion) external;
     function setTimestamp(uint256 id, uint64 timestamp) external;
     function setChainTeam(address newChainTeam) external;
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -34,12 +34,13 @@ interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBa
 
     function chainTeam() external view returns (address);
     function scheduleId() external view returns (bytes32);
+    function scheduleId(uint256 id) external view returns (bytes32);
     function minimumProtocolVersion() external view returns (uint256);
 
-    function getSchedule() external view returns (Upgrade[] memory schedule_);
+    function getSchedule() external view returns (Upgrade[] memory);
 
-    function initialize() external;
-    function registerUpgrade() external returns (uint256 id);
+    function initialize(address _chainTeam) external;
+    function registerUpgrade() external returns (uint256);
     function setMinimumProtocolVersion(uint256 protocolVersion) external;
     function setTimestamp(uint256 id, uint64 timestamp) external;
     function setChainTeam(address newChainTeam) external;
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -7,8 +7,8 @@ import { ISemver } from "interfaces/universal/ISemver.sol";
 /// @notice Interface for the ProtocolVersions upgrade schedule contract.
 interface IProtocolVersions is ISemver {
     struct Upgrade {
-        string  name;
-        uint64  timestamp;
+        string name;
+        uint64 timestamp;
         uint256 protocolVersion;
         bytes32 scheduleId;
     }
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -29,6 +29,7 @@ interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBa
     event TimestampSet(uint256 indexed id, uint256 timestamp);
     event ScheduleIdUpdated(bytes32 indexed newScheduleId, uint256 indexed blockNumber);
     event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
+    event Initialized(uint8 version);
 
     function MIN_NOTICE() external view returns (uint64);
 
@@ -36,7 +37,7 @@ interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBa
     function scheduleId() external view returns (bytes32);
     function minimumProtocolVersion() external view returns (uint256);
 
-    function getSchedule() external view returns (Upgrade[] memory);
+    function getSchedule() external view returns (Upgrade[] memory schedule_);
 
     function initialize(uint256 _l2ChainId) external;
     function registerUpgrade() external returns (uint256 id);
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -2,30 +2,30 @@
 pragma solidity ^0.8.0;
 
 import { ISemver } from "interfaces/universal/ISemver.sol";
+import { IProxyAdminOwnedBase } from "interfaces/L1/IProxyAdminOwnedBase.sol";
+import { IReinitializableBase } from "interfaces/universal/IReinitializableBase.sol";
 
 /// @title IProtocolVersions
 /// @notice Interface for the ProtocolVersions upgrade schedule contract.
-interface IProtocolVersions is ISemver {
+interface IProtocolVersions is IProxyAdminOwnedBase, ISemver, IReinitializableBase {
     struct Upgrade {
-        string name;
+        uint256 id;
         uint64 timestamp;
         bytes32 scheduleId;
     }
 
     error ProtocolVersions_InvalidL2ChainId();
-    error ProtocolVersions_UnknownUpgradeName(string upgradeId);
-    error ProtocolVersions_UpgradeAlreadyRegistered(bytes32 key);
-    error ProtocolVersions_InvalidUpgradeId();
+    error ProtocolVersions_UnknownUpgrade(uint256 id);
     error ProtocolVersions_InvalidProtocolVersion();
-    error ProtocolVersions_ActivationAlreadyPassed(bytes32 key, uint64 activationTimestamp);
-    error ProtocolVersions_ActivationTimestampInPast(uint64 timestamp);
+    error ProtocolVersions_ActivationAlreadyPassed(uint256 id, uint64 activationTimestamp);
     error ProtocolVersions_NotChainTeam();
-    error ProtocolVersions_NotScheduled(bytes32 key);
+    error ProtocolVersions_NotScheduled(uint256 id);
     error ProtocolVersions_DelayMustBeLater(uint64 currentTimestamp, uint64 newTimestamp);
 
-    event UpgradeRegistered(bytes32 indexed key, uint256 indexed index, string upgradeId, uint256 protocolVersion);
-    event TimestampSet(bytes32 indexed key, uint256 timestamp);
-
+    event UpgradeRegistered(uint256 indexed id, uint256 protocolVersion);
+    event LatestProtocolVersionUpdated(uint256 indexed protocolVersion);
+    event TimestampSet(uint256 indexed id, uint256 timestamp);
+    event ScheduleIdUpdated(bytes32 indexed newScheduleId, uint256 indexed blockNumber);
     event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
 
     function MIN_NOTICE() external view returns (uint64);
@@ -34,12 +34,14 @@ interface IProtocolVersions is ISemver {
     function scheduleId() external view returns (bytes32);
     function latestProtocolVersion() external view returns (uint256);
 
-    function getTimestamp(string calldata upgradeId) external view returns (uint256);
-
     function getSchedule() external view returns (Upgrade[] memory);
 
-    function registerUpgrade(string calldata upgradeId, uint256 protocolVersion) external;
-    function setTimestamp(string calldata upgradeId, uint64 timestamp) external;
+    function initialize(uint256 _l2ChainId) external;
+    function registerUpgrade(uint256 protocolVersion) external returns (uint256 id);
+    function setLatestProtocolVersion(uint256 protocolVersion) external;
+    function setTimestamp(uint256 id, uint64 timestamp) external;
     function setChainTeam(address newChainTeam) external;
-    function delayTimestamp(string calldata upgradeId, uint64 newTimestamp) external;
+    function delayTimestamp(uint256 id, uint64 newTimestamp) external;
+
+    function __constructor__() external;
 }
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { ISemver } from "interfaces/universal/ISemver.sol";
+
+interface IProtocolVersions is ISemver {
+    function l2ChainId() external view returns (uint256);
+    function chainTeam() external view returns (address);
+    function scheduleId() external view returns (bytes32);
+    function lastUpdatedAtBlock() external view returns (uint256);
+    function upgradeCount() external view returns (uint256);
+    function upgradeIdAt(uint256 index) external view returns (bytes32);
+    function upgradeIds() external view returns (bytes32[] memory);
+    function getTimestamp(string calldata upgradeId) external view returns (uint256);
+    function getProtocolVersion(string calldata upgradeId) external view returns (uint256);
+
+    function registerUpgrade(string calldata upgradeId, uint256 protocolVersion) external;
+    function setTimestamp(string calldata upgradeId, uint64 timestamp) external;
+    function setChainTeam(address newChainTeam) external;
+    function delayTimestamp(string calldata upgradeId, uint64 newTimestamp) external;
+}
--- a/interfaces/L1/IProtocolVersions.sol
+++ b/interfaces/L1/IProtocolVersions.sol
@@ -3,7 +3,27 @@ pragma solidity ^0.8.0;
 
 import { ISemver } from "interfaces/universal/ISemver.sol";
 
+/// @title IProtocolVersions
+/// @notice Interface for the ProtocolVersions upgrade schedule contract.
 interface IProtocolVersions is ISemver {
+    error ProtocolVersions_ZeroOwner();
+    error ProtocolVersions_InvalidL2ChainId();
+    error ProtocolVersions_UnknownUpgradeName(string upgradeId);
+    error ProtocolVersions_UpgradeAlreadyRegistered(bytes32 key);
+    error ProtocolVersions_InvalidUpgradeId();
+    error ProtocolVersions_InvalidProtocolVersion();
+    error ProtocolVersions_ActivationAlreadyPassed(bytes32 key, uint64 activationTimestamp);
+    error ProtocolVersions_ActivationTimestampInPast(uint64 timestamp);
+    error ProtocolVersions_NotChainTeam();
+    error ProtocolVersions_NotScheduled(bytes32 key);
+    error ProtocolVersions_DelayMustBeLater(uint64 currentTimestamp, uint64 newTimestamp);
+
+    event UpgradeRegistered(bytes32 indexed key, uint256 indexed index, string upgradeId, uint256 protocolVersion);
+    event TimestampSet(bytes32 indexed key, uint256 timestamp);
+    event ScheduleIdUpdated(bytes32 indexed oldScheduleId, bytes32 indexed newScheduleId, uint256 indexed blockNumber);
+    event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
+
+    function MIN_NOTICE() external view returns (uint64);
     function l2ChainId() external view returns (uint256);
     function chainTeam() external view returns (address);
     function scheduleId() external view returns (bytes32);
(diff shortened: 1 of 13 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+        // The upgrade must already have a scheduled activation to delay.
+        if (current == 0) revert ProtocolVersions_NotScheduled(key);
+        // Cannot modify an activation that has already passed.
+        if (uint64(block.timestamp) >= current) {
+            revert ProtocolVersions_ActivationAlreadyPassed(key, current);
+        }
+        // The role can only push the activation later, never earlier or to the same time.
+        if (newTimestamp <= current) revert ProtocolVersions_DelayMustBeLater(current, newTimestamp);
+
+        _timestamps[key] = newTimestamp;
```
jackchuma: Enforce notice window on chain-team delays. When a scheduled upgrade is already inside the one-hour notice window, this path accepts `newTimestamp` values that are only slightly later than `current` (for example, now+5m -> now+6m). That changes the canonical `scheduleId` without the `MIN_NOTICE` that `setTimestamp` enforces for every non-zero timestamp and that the contract documents as required for modifications, so `chainTeam` can bypass the notice window. Apply the same `newTimestamp < block.timestamp + MIN_NOTICE` guard here before writing the new timestamp

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+
+    /// @notice The L2 chain ID whose schedule this contract commits to.
+    uint256 public immutable l2ChainId;
+
+    /// @notice Hash chain seed: keccak256(abi.encode(l2ChainId, address(this))). Cached to avoid
+    ///         recomputing on every _refreshScheduleId call.
+    bytes32 internal immutable _seed;
+
+    /// @notice The latest block number in which `scheduleId` changed.
+    uint256 public lastUpdatedAtBlock;
```
jackchuma: Why is this necessary in storage?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+contract ProtocolVersions is Ownable, IProtocolVersions {
+    /// @notice Semantic version.
+    /// @custom:semver 1.0.0
+    string public constant version = "1.0.0";
+
+    /// @notice Minimum notice period required when scheduling or modifying an activation timestamp.
+    uint64 public constant MIN_NOTICE = 1 hours;
+
+    /// @notice The L2 chain ID whose schedule this contract commits to.
+    uint256 public immutable l2ChainId;
```
jackchuma: If l2ChainId is only needed to derive `_seed`, we should remove it as an immutable variable

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+
+    /// @notice Hash chain seed: keccak256(abi.encode(l2ChainId, address(this))). Cached to avoid
+    ///         recomputing on every _refreshScheduleId call.
+    bytes32 internal immutable _seed;
+
+    /// @notice The latest block number in which `scheduleId` changed.
+    uint256 public lastUpdatedAtBlock;
+
+    /// @notice The current schedule commitment hash, extended on every timestamp change.
+    bytes32 internal _scheduleId;
```
jackchuma: Why is this necessary in storage?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+    ///         j..n-1 links rather than the entire chain.
+    mapping(bytes32 => bytes32) internal _upgradeScheduleId;
+
+    /// @notice Activation timestamp for each registered upgrade key (0 = not scheduled).
+    mapping(bytes32 => uint64) internal _timestamps;
+
+    /// @notice Protocol version for each upgrade, chosen by the owner and set at registration via
+    ///         `registerUpgrade`. Immutable once registered. Non-zero for registered upgrades;
+    ///         zero for unregistered keys (serves as the existence sentinel).
+    mapping(bytes32 => uint256) internal _protocolVersions;
```
jackchuma: Isn't protocol version already implied by upgrade Id? I don't think this mapping is needed

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+// Contracts
+import { Ownable } from "lib/openzeppelin-contracts/contracts/access/Ownable.sol";
```
jackchuma: Nit: the codebase standard is to use [ProxyAdminOwnedBase](https://github.com/base/contracts/blob/main/src/universal/ProxyAdminOwnedBase.sol) instead of `Ownable`

interfaces/L1/IProtocolVersions.sol
```diff
@@ -0,0 +1,50 @@
...
+    event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
+
+    function MIN_NOTICE() external view returns (uint64);
+    function l2ChainId() external view returns (uint256);
+    function chainTeam() external view returns (address);
+    function scheduleId() external view returns (bytes32);
+    function lastUpdatedAtBlock() external view returns (uint256);
+    function upgradeCount() external view returns (uint256);
+    function upgradeIdAt(uint256 index) external view returns (bytes32);
+    function upgradeIds() external view returns (bytes32[] memory);
```
jackchuma: These registry views duplicate what `getSchedule()` already exposes, and I do not see non-test callers for the extra count/index/list API. Keeping only `getSchedule()` gives consumers one way to read the registry and avoids locking in three more public methods. Suggest deleting `upgradeCount`, `upgradeIdAt`, and `upgradeIds` unless there is a known external consumer for them

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+        l2ChainId = _l2ChainId;
+        _seed = keccak256(abi.encode(_l2ChainId, address(this)));
+
+        _scheduleId = _seed;
+        lastUpdatedAtBlock = block.number;
+        emit ScheduleIdUpdated(bytes32(0), _scheduleId, block.number);
+    }
+
+    /// @notice Restricts a call to the appointed chainTeam role.
+    modifier onlyChainTeam() {
```
jackchuma: `onlyChainTeam` has one caller, so the modifier adds a layer without reuse. Since the auth check is a single line, inlining `if (msg.sender != chainTeam) revert ProtocolVersions_NotChainTeam();` inside `delayTimestamp` would be shorter and easier to scan

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+            });
+        }
+    }
+
+    /// @notice Registers a new upgrade by upgradeId with its protocol version. Owner only.
+    /// @dev The upgradeId is packed into a bytes32 key and `protocolVersion` is recorded for it.
+    ///      Registration extends the scheduleId chain with the new (key, timestamp=0) link.
+    /// @param upgradeId       Human-readable upgrade name (1–32 bytes, e.g. "canyon").
+    /// @param protocolVersion Owner-assigned protocol version number for this upgrade (must be non-zero).
+    function registerUpgrade(string calldata upgradeId, uint256 protocolVersion) external onlyOwner {
```
jackchuma: `registerUpgrade` only forwards to `_registerUpgrade`, and `_registerUpgrade` has no other caller. This split does not buy much right now. Suggest inlining `_registerUpgrade` into `registerUpgrade` until there is a second path that needs the same registration logic

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,298 @@
...
+
+        _timestamps[key] = newTimestamp;
+        emit TimestampSet(key, newTimestamp);
+        _refreshScheduleId(key);
+    }
+
+    /// @dev Validates and applies a timestamp update for a registered upgrade key.
+    ///      Returns true if the timestamp changed (caller should then call _refreshScheduleId),
+    ///      false if unchanged (no-op, scheduleId must not be extended).
+    function _applyTimestamp(bytes32 key, uint64 timestamp) internal returns (bool) {
```
jackchuma: `_applyTimestamp` is only used by `setTimestamp`, and the bool return mainly exists to coordinate that one caller with `_refreshScheduleId`. Inlining the validation/update into `setTimestamp` would remove the helper and make the no-op path explicit where the timestamp is being changed

jackchuma (commented):
I think this is a great start, however `ProtocolVersions` seems a bit over-engineered. How simple can we make it? We really only need: 

- mapping of arbitrary upgradeId `bytes32` => upgrade timestamp
- We could probably also include another mapping to represent the upgrade schedule as a linked list instead of an array which will simplify lookups (gnosis safes are implemented this way)
- mapping of upgradeId `bytes32` => scheduleId `bytes32`

src/L1/ProtocolVersions.sol
PelleKrab: Extra fn, not needed.

src/L1/ProtocolVersions.sol
PelleKrab: We can use the tip of the hsah chain instead.

src/L1/ProtocolVersions.sol
PelleKrab: After the refactor there was some duplication, so i reintroduced this fn as `_writeTimestamp()`.

test/vendor/Initializable.t.sol
```diff
@@ -256,8 +256,9 @@ contract Initializer_Test is CommonTest {
         excludes[j++] = "src/periphery/*";
         // L2 contract initialization is tested in Predeploys.t.sol
         excludes[j++] = "src/L2/*";
-        // Contract is not deployed as part of the standard deployment script.
+        // Contracts not deployed as part of the standard deployment script.
         excludes[j++] = "src/L1/BalanceTracker.sol";
+        excludes[j++] = "src/L1/ProtocolVersions.sol";
```
jackchuma: Nit: please add ProtocolVersions to the standard deployment script

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+    /// @custom:semver 1.0.0
+    string public constant version = "1.0.0";
+
+    /// @notice Minimum notice period required when scheduling or modifying an activation timestamp.
+    uint64 public constant MIN_NOTICE = 1 hours;
+
+    /// @notice Hash chain seed: keccak256(abi.encode(l2ChainId, address(this))). Set once in
+    ///         `initialize` (run through the proxy, so `address(this)` is the proxy address) and
+    ///         cached in storage to avoid recomputing on every _refreshScheduleId call.
+    bytes32 internal _seed;
```
jackchuma: Can we eliminate this and just inline / use bytes32(0) as the seed below?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+        _assertOnlyProxyAdminOwner();
+        _;
+    }
+
+    /// @notice Returns the canonical schedule commitment.
+    /// @return The current scheduleId hash.
+    function scheduleId() external view returns (bytes32) {
+        if (_seed == 0) revert ProtocolVersions_NotInitialized();
+        uint256 n = _upgradeScheduleId.length;
+        return n == 0 ? _seed : _upgradeScheduleId[n - 1];
```
jackchuma: We can remove this if / else by initializing the `_upgradeScheduleId` array in the `initialize` function above to have `_seed` as its only element

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+import { Initializable } from "lib/openzeppelin-contracts/contracts/proxy/utils/Initializable.sol";
+
+// Interfaces
+import { ISemver } from "interfaces/universal/ISemver.sol";
+
+/// @title ProtocolVersions
+/// @notice Security Council-controlled upgrade activation schedule contract.
+/// @dev Maintains an ordered registry of upgrades and their L2 activation timestamps.
+///      Each upgrade is identified by an ascending numeric `id` equal to its registration
+///      index (0, 1, 2, ...). Human-readable names are intentionally kept off-chain: a client
```
jackchuma: Base branding nit: should use "offchain" instead of "off-chain"

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+
+    /// @notice Disables initializers on the implementation so it can only be used behind a proxy.
+    constructor() ReinitializableBase(1) {
+        _disableInitializers();
+    }
+
+    /// @notice Initializes the registry, computing the hash chain seed from the proxy context and
+    ///         emitting the initial scheduleId. Callable only by the ProxyAdmin or its owner.
+    /// @param _l2ChainId L2 chain ID whose schedule is being committed.
+    function initialize(uint256 _l2ChainId) external reinitializer(initVersion()) {
```
jackchuma: Can we set `chainTeam` in the initialize function too?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+    /// @notice Emitted when an upgrade's activation timestamp is set, cleared, or delayed.
+    event TimestampSet(uint256 indexed id, uint256 timestamp);
+    /// @notice Emitted when the schedule commitment changes.
+    event ScheduleIdUpdated(bytes32 indexed newScheduleId, uint256 indexed blockNumber);
+    /// @notice Emitted when the chainTeam role changes.
+    event ChainTeamUpdated(address indexed previousChainTeam, address indexed newChainTeam);
+
+    /// @notice Semantic version.
+    /// @custom:semver 1.0.0
+    string public constant version = "1.0.0";
```
jackchuma: Nit: it's against our style guide to have a lowercased constant - this is why most other contracts return the version from a public view function

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+        uint256 n = _upgradeScheduleId.length;
+        return n == 0 ? _seed : _upgradeScheduleId[n - 1];
+    }
+
+    /// @notice Returns the full ordered schedule: every registered upgrade with its current
+    ///         activation timestamp and cumulative schedule hash. The array index equals the
+    ///         upgrade `id`; names are resolved off-chain.
+    /// @dev Calling via eth_call is gas-free; no transaction is submitted.
+    /// @return schedule_ Ordered array of Upgrade structs, one per registered upgrade.
+    function getSchedule() external view returns (Upgrade[] memory schedule_) {
```
jackchuma: Nit: prefer explicit return statements at the bottom of the function instead of named variable returns

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+
+    /// @notice Restricts a call to the ProxyAdmin owner.
+    modifier onlyProxyAdminOwner() {
+        _assertOnlyProxyAdminOwner();
+        _;
+    }
+
+    /// @notice Returns the canonical schedule commitment.
+    /// @return The current scheduleId hash.
+    function scheduleId() external view returns (bytes32) {
```
jackchuma: We probably also want a `scheduleId(uint256)` function as well so arbitrary scheduleIds can be queried

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+///      Proof journals bind to `scheduleId`, pinning every proof in a dispute game to the schedule
+///      that was in effect at the game's L1 origin block.
+///
+///      The contract is deployed behind an OP proxy: the implementation constructor disables
+///      initializers, and `initialize` (run through the proxy) computes and stores `_seed` from the
+///      proxy's own `address(this)`, so the schedule commitment is bound to the proxy address that
+///      callers and off-chain consumers treat as the registry.
+contract ProtocolVersions is ProxyAdminOwnedBase, Initializable, ReinitializableBase, ISemver {
+    /// @notice A registered upgrade's id, current activation timestamp, and cumulative schedule hash.
+    struct Upgrade {
```
jackchuma: I think `Upgrade.id` and per-entry `Upgrade.scheduleId` are extra API surface here. `id` is already the array index, and consumers can reproduce the hash chain from the timestamps plus seed if they need intermediate links. Returning just `uint64[]` timestamps from `getSchedule()` would make the contract ABI and storage model smaller

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+    error ProtocolVersions_InsufficientNotice(uint64 timestamp);
+
+    /// @notice Emitted when a new upgrade is registered.
+    event UpgradeRegistered(uint256 indexed id);
+    /// @notice Emitted when the minimum protocol version clients must run is updated.
+    event MinimumProtocolVersionUpdated(uint256 indexed protocolVersion);
+    /// @notice Emitted when an upgrade's activation timestamp is set, cleared, or delayed.
+    event TimestampSet(uint256 indexed id, uint256 timestamp);
+    /// @notice Emitted when the schedule commitment changes.
+    event ScheduleIdUpdated(bytes32 indexed newScheduleId, uint256 indexed blockNumber);
```
jackchuma: `ScheduleIdUpdated` includes `blockNumber`, but logs already carry the block number in event metadata. Dropping this parameter keeps the event smaller without losing information

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+        _assertOnlyProxyAdminOrProxyAdminOwner();
+        if (_l2ChainId == 0) revert ProtocolVersions_InvalidL2ChainId();
+
+        // address(this) is the proxy address, so the seed binds the schedule to the proxy registry.
+        _seed = keccak256(abi.encode(_l2ChainId, address(this)));
+
+        emit ScheduleIdUpdated(_seed, block.number);
+    }
+
+    /// @notice Restricts a call to the ProxyAdmin owner.
```
jackchuma: `onlyProxyAdminOwner` is a local modifier with one statement and only a few call sites. Calling `_assertOnlyProxyAdminOwner()` inline, like nearby contracts do, removes a layer without changing behavior

test/L1/ProtocolVersions.t.sol
```diff
@@ -0,0 +1,552 @@
...
+        vm.prank(_owner);
+        protocolVersions.setTimestamp(CANYON, ts_);
+    }
+}
+
+/// @title ProtocolVersions_Initialize_Test
+/// @notice Test contract for the ProtocolVersions initializer.
+contract ProtocolVersions_Initialize_Test is ProtocolVersions_TestInit {
+    /// @notice Tests that initialization sets the correct initial state.
+    function test_initialize_setsInitialState_succeeds() external view {
```
jackchuma, line 54: `test_initialize_setsInitialState_succeeds` checks that `scheduleId()` is nonzero, but the next test asserts the exact expected seed. The exact assertion already covers nonzero, so this test can go

test/L1/ProtocolVersions.t.sol
```diff
@@ -0,0 +1,552 @@
...
+        protocolVersions.setChainTeam(address(0));
+
+        assertEq(protocolVersions.chainTeam(), address(0));
+    }
+}
+
+/// @title ProtocolVersions_Uncategorized_Test
+/// @notice Test contract for view functions and the upgrade registry.
+contract ProtocolVersions_Uncategorized_Test is ProtocolVersions_TestInit {
+    /// @notice Tests that the registry starts empty.
```
jackchuma: `test_registry_startsEmpty_succeeds` and `test_getSchedule_empty_succeeds` assert the same empty schedule behavior. Keep one test and delete the duplicate

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+        if (msg.sender != chainTeam) revert ProtocolVersions_NotChainTeam();
+        _assertRegistered(id);
+        uint64 current = _timestamps[id];
+
+        // The upgrade must already have a scheduled activation to delay.
+        if (current == 0) revert ProtocolVersions_NotScheduled(id);
+        // Cannot delay an activation that has already passed.
+        if (uint64(block.timestamp) >= current) revert ProtocolVersions_ActivationAlreadyPassed(id, current);
+        // The role can only push the activation later, never to the same time or earlier.
+        if (newTimestamp <= current) revert ProtocolVersions_DelayMustBeLater(current, newTimestamp);
```
jackchuma, line 209: I'm debating if we need this. Since the following if statement is enforcing a min notice, maybe it's ok to allow `newTimestamp` to be <= `current`. What do you think?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,263 @@
...
+        for (uint256 i = 0; i < n; i++) {
+            schedule_[i] = Upgrade({ id: i, timestamp: _timestamps[i], scheduleId: _upgradeScheduleId[i] });
+        }
+    }
+
+    /// @notice Registers a new upgrade, assigning it the next ascending id. Owner only.
+    /// @dev The upgrade starts unscheduled (timestamp 0). Registration extends the scheduleId
+    ///      chain with the new (id, timestamp=0) link.
+    /// @return id The ascending id assigned to the newly registered upgrade.
+    function registerUpgrade() external onlyProxyAdminOwner returns (uint256 id) {
```
jackchuma: Can we set the timestamp and minimum protocol version here too? I don't see the point of refreshing schedule ID based off a 0 timestamp

src/L1/ProtocolVersions.sol
jackchuma: This goes for a handful of functions below as well

src/L1/ProtocolVersions.sol
PelleKrab: We should be good to remove it from a technical standpoint. Although I think it is best to keep it, as SC should be the only one who moves the timestamps back based on prev. discussions about what chain team should be able to control.

scripts/deploy/DeployConfig.s.sol
```diff
@@ -12,6 +12,7 @@ contract DeployConfig is Script {
 
     address public baseFeeVaultRecipient;
     address public batchSenderAddress;
+    address public chainTeam;
```
jackchuma: This is actually the same value as the `superchainConfigIncidentResponder` config below. Can we combine?

src/L1/ProtocolVersions.sol
```diff
@@ -0,0 +1,286 @@
...
+        _assertOnlyProxyAdminOwner();
+        _assertRegistered(id);
+        uint64 current = _timestamps[id];
+        if (current != 0 && uint64(block.timestamp) >= current) {
+            revert ProtocolVersions_ActivationAlreadyPassed(id, current);
+        }
+        if (timestamp != 0 && timestamp < uint64(block.timestamp) + MIN_NOTICE) {
+            revert ProtocolVersions_InsufficientNotice(timestamp);
+        }
+        if (current == timestamp) return;
```
jackchuma: Maybe move this if statement before the other 2 so no-op txns succeed

scripts/deploy/SystemDeploy.s.sol
```diff
@@ -619,6 +625,13 @@ contract SystemDeploy is Script {
             _impls.anchorStateRegistryImpl,
             _encodeAnchorStateRegistryInitializer(_input, _output)
         );
+
+        _upgradeToAndCall(
```
jackchuma, line 635: Upgrade ProtocolVersions proxies too. For chains deployed with this new `ProtocolVersionsProxy`, a later `SystemDeploy.upgrade` with a newer `protocolVersionsImpl` will leave this proxy on the old implementation: `_assertValidImplementations` now requires the impl, but `_upgradeOPChain` never discovers or calls `_upgradeTo` for the ProtocolVersions proxy. Please add ProtocolVersions to the upgrade input/discovery path, or future standard upgrades will silently skip it while upgrading the rest of the system

jackchuma (approved):

PR #358, opened 2026-07-08
feat(L2): add system address refunding to FeeDisburser

0xth4nh:
Closes CHAIN-4644

## What changed?

\`FeeDisburser\` now tops up a configurable set of L2 system addresses before bridging the remaining fees to L1. Previously, the full collected balance was always bridged.

Specifically:
- \`disburseFees()\` iterates over configured system addresses after vault withdrawals and before bridging. Each address is refilled up to its target balance via \`SafeCall.send\` (non-reverting — a failing recipient does not abort the rest).
- \`setSystemAddresses(address payable[], uint256[])\` added — ProxyAdmin-gated setter for the system address/target-balance config. Validates: ≤20 entries, equal-length arrays, no \`address(0)\`, no zero target. Passing empty arrays clears the config and disables refunding.
- \`nonReentrantDisbursement\` modifier added — custom reentrancy guard, not OZ's \`ReentrancyGuardUpgradeable\`. OZ cannot be used here because \`FeeDisburser\` is an upgradeable proxy with existing deployed storage; inheriting \`ReentrancyGuardUpgradeable\` would prepend its \`_status\` slot to the layout, shifting every existing variable and corrupting proxy storage on upgrade. Instead, \`_disburseFeesEntered\` is appended after the existing storage slots. It uses 1/2 sentinels (not 0/1/0) so the slot stays nonzero after first use, keeping subsequent SSTOREs warm; \`0\` (uninitialized proxy slot) is treated as not-entered.
- \`FeeDisburser\` now inherits \`ProxyAdminOwnedBase\` (provides \`_assertOnlyProxyAdmin\` — reads the EIP-1967 admin slot directly, compatible with the raw L2-alias admin used in production without requiring a \`ProxyAdmin\` contract).
- New events: \`ProcessedFunds(systemAddress, success, balanceNeeded, balanceSent)\`, \`SystemAddressesUpdated(systemAddressCount)\`.
- New errors: \`ArrayLengthMismatch\`, \`TooManySystemAddresses\`, \`ZeroTargetBalance\`, \`ReentrantCall\`.
- If refunds consume the entire balance, \`FeesDisbursed(time, 0, 0)\` is emitted and \`lastDisbursementTime\` is still updated (disbursement did occur, just nothing left to bridge).
- Version bumped \`1.0.0\` → \`1.1.0\`.

Mirrors the existing \`BalanceTracker\` pattern on L1 (\`src/L1/BalanceTracker.sol\`).

## Why?

L2 system addresses currently must be funded out-of-band, independently of the fee collection cycle. Routing a first-priority top-up through \`FeeDisburser\` (same contract that already holds all collected fees) eliminates that operational dependency and keeps the L1 and L2 funding patterns consistent.

## How to test?

```bash
forge test --match-path test/L2/FeeDisburser.t.sol -vvv
```

Tests cover:
1. **Full refund** — system address reaches target, remainder bridges to L1.
2. **Partial refund (insufficient funds)** — target exceeds available balance; system address receives all available funds, nothing bridged.
3. **Skip when already at target** — \`ProcessedFunds(addr, false, 0, 0)\` emitted, full balance bridges.
4. **Reverting recipient** — recipient reverts on receive; \`SafeCall.send\` catches it, \`success=false\` emitted, full balance still bridges.
5. **Reentrancy blocked** — system address calls back into \`disburseFees()\` on receive; inner call reverts with \`ReentrantCall()\`, caught by \`SafeCall\`, outer call completes normally.
6. **Multiple system addresses** — correct ordering and cumulative deduction from fee balance.
7. **Clear config** — passing empty arrays succeeds and disables refunding.
8. **\`setSystemAddresses\` access control** — non-ProxyAdmin caller reverts.

### DIFF
--- a/src/L2/FeeDisburser.sol
+++ b/src/L2/FeeDisburser.sol
@@ -7,12 +7,11 @@ import { ISemver } from "interfaces/universal/ISemver.sol";
 import { Predeploys } from "src/libraries/Predeploys.sol";
 import { SafeCall } from "src/libraries/SafeCall.sol";
 import { Initializable } from "src/vendor/Initializable.sol";
-import { ReentrancyGuard } from "lib/solady/src/utils/ReentrancyGuard.sol";
 
 /// @custom:proxied true
 /// @title FeeDisburser
 /// @notice Withdraws funds from system FeeVault contracts and bridges to L1.
-contract FeeDisburser is Initializable, ReentrancyGuard, ISemver {
+contract FeeDisburser is Initializable, ISemver {
     ////////////////////////////////////////////////////////////////
     ///                     Constants
     ////////////////////////////////////////////////////////////////
@@ -44,6 +43,9 @@ contract FeeDisburser is Initializable, ReentrancyGuard, ISemver {
     /// @notice Previously tracked the aggregate net fee revenue (sum of sequencer and base fees).
     ///         This variable is deprecated and its value should not be relied upon.
     uint256 public netFeeRevenue;
+
+    /// @notice Reentrancy guard status for disburseFees.
+    uint256 private _disburseFeesEntered;
 
     /// @notice The L2 system addresses being funded.
     address payable[] public systemAddresses;
@@ -110,6 +112,23 @@ contract FeeDisburser is Initializable, ReentrancyGuard, ISemver {
     /// @notice Thrown when a system address target balance is zero.
     error ZeroTargetBalance();
 
+    /// @notice Thrown when disburseFees is reentered.
+    error ReentrantCall();
+
+    ////////////////////////////////////////////////////////////////
+    ///                       Modifiers
+    ////////////////////////////////////////////////////////////////
+
+    /// @notice Prevents reentrancy into disburseFees while preserving the existing storage layout.
+    ///         Uses 1/2 sentinel values so the slot stays nonzero after first use, keeping
+    ///         subsequent SSTOREs warm. Uninitialized (0) is treated as not-entered.
+    modifier nonReentrantDisbursement() {
+        if (_disburseFeesEntered == 2) revert ReentrantCall();
+        _disburseFeesEntered = 2;
+        _;
+        _disburseFeesEntered = 1;
+    }
+
     ////////////////////////////////////////////////////////////////
     ///                     Constructor
     ////////////////////////////////////////////////////////////////
@@ -131,7 +150,7 @@ contract FeeDisburser is Initializable, ReentrancyGuard, ISemver {
     ////////////////////////////////////////////////////////////////
 
     /// @notice Withdraws funds from FeeVaults and bridges to L1.
-    function disburseFees() external virtual nonReentrant {
+    function disburseFees() external virtual nonReentrantDisbursement {
         if (block.timestamp < lastDisbursementTime + FEE_DISBURSEMENT_INTERVAL) revert IntervalNotReached();
 
         // Sequencer, base, and L1 FeeVaults will withdraw fees to the FeeDisburser contract.
--- a/src/L2/FeeDisburser.sol
+++ b/src/L2/FeeDisburser.sol
@@ -78,7 +78,8 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
     /// @param systemAddress The system address being funded.
     /// @param success       A boolean denoting whether a fund send occurred and its success or failure.
     /// @param balanceNeeded The amount of funds the given system address needs to reach its target balance.
-    /// @param balanceSent   The amount of funds sent to the system address.
+    /// @param balanceSent   The amount of funds attempted to be sent. When success is false, the
+    ///                      recipient rejected the transfer and no funds were actually received.
     event ProcessedFunds(
         address indexed systemAddress, bool indexed success, uint256 balanceNeeded, uint256 balanceSent
     );
@@ -110,9 +111,6 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
     /// @notice Thrown when system address and target balance array lengths do not match.
     error ArrayLengthMismatch();
 
-    /// @notice Thrown when system address configuration is empty.
-    error EmptySystemAddresses();
-
     /// @notice Thrown when system address configuration exceeds the maximum length.
     error TooManySystemAddresses();
 
@@ -127,11 +125,13 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
     ////////////////////////////////////////////////////////////////
 
     /// @notice Prevents reentrancy into disburseFees while preserving the existing storage layout.
+    ///         Uses 1/2 sentinel values so the slot stays nonzero after first use, keeping
+    ///         subsequent SSTOREs warm. Uninitialized (0) is treated as not-entered.
     modifier nonReentrantDisbursement() {
-        if (_disburseFeesEntered != 0) revert ReentrantCall();
-        _disburseFeesEntered = 1;
+        if (_disburseFeesEntered == 2) revert ReentrantCall();
+        _disburseFeesEntered = 2;
         _;
-        _disburseFeesEntered = 0;
+        _disburseFeesEntered = 1;
     }
 
     ////////////////////////////////////////////////////////////////
@@ -199,8 +199,9 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
     }
 
     /// @notice Configures the L2 system addresses to refund and their target balances.
+    ///         Passing empty arrays clears the configuration and disables refunding.
     ///
-    /// @dev Callable only by the ProxyAdmin owner.
+    /// @dev Callable only by the ProxyAdmin.
     ///
     /// @param systemAddresses_ The system addresses being funded.
     /// @param targetBalances_  The target balances for system addresses.
@@ -214,7 +215,6 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
         _assertOnlyProxyAdmin();
 
         uint256 systemAddressesLength = systemAddresses_.length;
-        if (systemAddressesLength == 0) revert EmptySystemAddresses();
         if (systemAddressesLength > MAX_SYSTEM_ADDRESS_COUNT) revert TooManySystemAddresses();
         if (systemAddressesLength != targetBalances_.length) revert ArrayLengthMismatch();
 
@@ -250,7 +250,7 @@ contract FeeDisburser is ProxyAdminOwnedBase, ISemver {
     ///
     /// @param systemAddress The system address being funded.
     /// @param targetBalance The target balance for the system address being funded.
-    function _refillBalanceIfNeeded(address systemAddress, uint256 targetBalance) internal virtual {
+    function _refillBalanceIfNeeded(address systemAddress, uint256 targetBalance) internal {
         uint256 systemAddressBalance = systemAddress.balance;
         if (systemAddressBalance >= targetBalance) {
             emit ProcessedFunds({ systemAddress: systemAddress, success: false, balanceNeeded: 0, balanceSent: 0 });
--- a/src/L2/FeeDisburser.sol
+++ b/src/L2/FeeDisburser.sol
@@ -7,12 +7,12 @@ import { ISemver } from "interfaces/universal/ISemver.sol";
 import { Predeploys } from "src/libraries/Predeploys.sol";
 import { SafeCall } from "src/libraries/SafeCall.sol";
 import { Initializable } from "src/vendor/Initializable.sol";
-import { ProxyAdminOwnedBase } from "src/universal/ProxyAdminOwnedBase.sol";
+import { ReentrancyGuard } from "lib/solady/src/utils/ReentrancyGuard.sol";
 
 /// @custom:proxied true
 /// @title FeeDisburser
 /// @notice Withdraws funds from system FeeVault contracts and bridges to L1.
-contract FeeDisburser is Initializable, ProxyAdminOwnedBase, ISemver {
+contract FeeDisburser is Initializable, ReentrancyGuard, ISemver {
     ////////////////////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

src/L2/FeeDisburser.sol
```diff
@@ -39,6 +44,15 @@ contract FeeDisburser is ISemver {
     ///         This variable is deprecated and its value should not be relied upon.
     uint256 public netFeeRevenue;
 
+    /// @notice Reentrancy guard status for disburseFees.
+    uint256 private _disburseFeesEntered;
```
jackchuma, line 49: **[P1] Regenerate FeeDisburser snapshots**

Because this change adds storage, ABI entries, and a new semver for FeeDisburser, the generated snapshots are now stale: `snapshots/semver-lock.json` still has the old FeeDisburser hashes, and the ABI/storage layout snapshots do not include these new members. CI runs `just semver-lock` and will fail with a dirty `snapshots/semver-lock.json` until the snapshots are regenerated.

src/L2/FeeDisburser.sol
```diff
@@ -117,24 +173,98 @@ contract FeeDisburser is ISemver {
...
+    /// @param systemAddresses_ The system addresses being funded.
+    /// @param targetBalances_  The target balances for system addresses.
+    function setSystemAddresses(
+        address payable[] memory systemAddresses_,
+        uint256[] memory targetBalances_
+    )
+        external
+        virtual
+    {
+        _assertOnlyProxyAdmin();
```
jackchuma: **[P2] Allow the proxy admin owner to update the config**

When FeeDisburser is called through the repo's transparent proxy, normal governance calls come from the ProxyAdmin owner and are delegated through the proxy, while the admin slot contains the ProxyAdmin contract itself. `_assertOnlyProxyAdmin()` rejects those owner calls, and `ProxyAdmin` has no generic call method other than `upgradeAndCall`, so after the upgrade this mutable configuration cannot be updated or cleared through the standard admin flow.

jackchuma (dismissed):
The patch leaves required generated snapshots stale, which will fail repository checks, and the new setter is not reachable through the standard proxy admin owner flow. These should be fixed before the change is considered correct.

src/L2/FeeDisburser.sol
```diff
@@ -117,24 +173,98 @@ contract FeeDisburser is ISemver {
...
+    }
+
+    /// @notice Configures the L2 system addresses to refund and their target balances.
+    ///         Passing empty arrays clears the configuration and disables refunding.
+    ///
+    /// @dev Callable only by the ProxyAdmin.
+    ///
+    /// @param systemAddresses_ The system addresses being funded.
+    /// @param targetBalances_  The target balances for system addresses.
+    function setSystemAddresses(
```
jackchuma: Would prefer for this to be an initialize function so this only gets set during upgrades. Similar to how BalanceTracker is set up

test/L2/FeeDisburser.t.sol
```diff
@@ -6,7 +6,27 @@ import { Test } from "lib/forge-std/src/Test.sol";
...
+contract ReentrantReceiver {
+    FeeDisburser public target;
+
+    constructor(FeeDisburser _target) {
+        target = _target;
+    }
+
+    receive() external payable {
+        target.disburseFees();
+    }
```
jackchuma: `ReentrantReceiver` stores the target only so `receive()` can call back into it. Since the sender of the ETH transfer is the `FeeDisburser`, the test contract can call `FeeDisburser(payable(msg.sender)).disburseFees()` directly and avoid the state variable plus constructor.

  Suggested simplification: make `ReentrantReceiver` stateless

test/L2/FeeDisburser.t.sol
```diff
@@ -6,7 +6,27 @@ import { Test } from "lib/forge-std/src/Test.sol";
...
+}
+
+contract RevertingReceiver {
+    receive() external payable {
+        revert();
+    }
+}
```
jackchuma: This reverting receiver contract exists only to make a payable call fail. Foundry can model that directly without another helper contract.

  Suggested simplification: use `vm.mockCallRevert(...)` for the recipient instead of deploying `RevertingReceiver`

test/L2/FeeDisburser.t.sol
```diff
@@ -768,4 +818,307 @@ contract FeeDisburserTest is Test {
...
+
+    function test_disburseFees_success_systemAddressRefunded() public {
+        uint256 feeAmount = 10 ether;
+        uint256 targetBal = 2 ether;
+
+        (address payable[] memory addrs, uint256[] memory balances) = _makeSingleConfig(SYSTEM_ADDR, targetBal);
+        _setSystemAddresses(addrs, balances);
+
+        _mockVaultWithdrawal(Predeploys.SEQUENCER_FEE_WALLET, feeAmount);
+        vm.deal(Predeploys.BASE_FEE_VAULT, 0);
```
jackchuma: Several new refund tests repeatedly set `BASE_FEE_VAULT` and `L1_FEE_VAULT` balances to zero even though each test starts from fresh state and those balances are already zero unless explicitly funded. The repeated setup makes the important part of each test harder to see.

  Suggested simplification: delete the repeated zeroing lines and only set balances for vaults that matter to the test

test/L2/FeeDisburser.t.sol
```diff
@@ -768,4 +818,307 @@ contract FeeDisburserTest is Test {
...
+        emit ProcessedFunds(SYSTEM_ADDR, false, 0, 0);
+
+        _expectFeesDisbursed(feeAmount);
+
+        feeDisburser.disburseFees();
+
+        assertEq(SYSTEM_ADDR.balance, targetBal);
+    }
+
+    function test_disburseFees_success_allFundsConsumedByRefund() public {
```
jackchuma: `test_disburseFees_success_allFundsConsumedByRefund` and `test_disburseFees_success_partialRefundDueToLowBalance` cover the same branch: the refund consumes all available FeeDisburser ETH, no bridge call is made, and `FeesDisbursed(0)` is emitted. The second test is the more general version because it also covers an address with a starting balance.

  Suggested simplification: keep the partial-refund test and delete the all-funds-consumed variant

src/L2/FeeDisburser.sol
0xth4nh: Fixed — snapshots regenerated.

src/L2/FeeDisburser.sol
0xth4nh: Fixed — replaced `setSystemAddresses` with `initialize(...) reinitializer(2)`. Config is now set atomically via `upgradeAndCall`, so access control is handled by the upgrade mechanism itself.

src/L2/FeeDisburser.sol
0xth4nh: Fixed — done as part of P2.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed — using `vm.mockCallRevert` instead.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed — deleted `allFundsConsumedByRefund`.

src/L2/FeeDisburser.sol
```diff
@@ -117,24 +168,92 @@ contract FeeDisburser is ISemver {
...
+    ///         Called via upgradeAndCall when upgrading to this version.
+    ///
+    /// @param systemAddresses_ The system addresses being funded.
+    /// @param targetBalances_  The target balances for system addresses.
+    function initialize(
+        address payable[] memory systemAddresses_,
+        uint256[] memory targetBalances_
+    )
+        external
+        reinitializer(2)
```
jackchuma, line 202: Restrict one-time fee recipient initialization. If this proxy is upgraded or deployed without executing this initializer in the same transaction, any account can call the public reinitializer first, permanently consume version 2, and set `systemAddresses`/`targetBalances` to addresses they control; subsequent `disburseFees` calls will top those addresses up before bridging fees. This should be limited to the ProxyAdmin/owner or otherwise made impossible to front-run

src/L2/FeeDisburser.sol
```diff
@@ -117,24 +168,92 @@ contract FeeDisburser is ISemver {
...
+
+        uint256 systemAddressesLength = systemAddresses.length;
+        for (uint256 i; i < systemAddressesLength;) {
+            _refillBalanceIfNeeded({ systemAddress: systemAddresses[i], targetBalance: targetBalances[i] });
+            unchecked {
+                i++;
+            }
+        }
+
+        uint256 bridgeBalance = address(this).balance;
```
jackchuma, line 179: nit: the zero-bridge case has its own early return, but it emits the same event shape as the normal path with `bridgeBalance == 0`. You can avoid the duplicated `FeesDisbursed` path by only guarding the bridge call:

```solidity
  uint256 bridgeBalance = address(this).balance;
  if (bridgeBalance != 0) {
      IL2StandardBridge(payable(Predeploys.L2_STANDARD_BRIDGE)).bridgeETHTo{ value: bridgeBalance }(
          L1_WALLET, WITHDRAWAL_MIN_GAS, bytes("")
      );
  }

  emit FeesDisbursed(disbursementTime, 0, bridgeBalance);
```

test/L2/FeeDisburser.t.sol
```diff
@@ -768,4 +792,220 @@ contract FeeDisburserTest is Test {
...
+    }
+
+    function test_initialize_revert_tooManySystemAddresses() public {
+        uint256 n = feeDisburser.MAX_SYSTEM_ADDRESS_COUNT() + 1;
+        address payable[] memory addrs = new address payable[](n);
+        uint256[] memory balances = new uint256[](n);
+        for (uint256 i; i < n; i++) {
+            addrs[i] = payable(address(uint160(0x6000 + i)));
+            balances[i] = 1 ether;
+        }
```
jackchuma: nit: this test expects `TooManySystemAddresses`, which happens before the initializer reads any array values. Filling every address and balance makes the test look like it depends on valid contents, but it only depends on length. These lines can be deleted

test/L2/FeeDisburser.t.sol
```diff
@@ -768,4 +792,220 @@ contract FeeDisburserTest is Test {
...
+
+        vm.expectRevert(FeeDisburser.TooManySystemAddresses.selector);
+        feeDisburser.initialize(addrs, balances);
+    }
+
+    function test_initialize_revert_arrayLengthMismatch() public {
```
jackchuma, line 867: Populating the arrays in this test is unnecessary since only the length mismatch is what is tested

src/L2/FeeDisburser.sol
0xth4nh: Fixed.

src/L2/FeeDisburser.sol
0xth4nh: Fixed.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed.

test/L2/FeeDisburser.t.sol
0xth4nh: Fixed.

src/L2/FeeDisburser.sol
```diff
@@ -117,24 +169,93 @@ contract FeeDisburser is ISemver {
             return;
         }
 
-        lastDisbursementTime = block.timestamp;
+        uint256 disbursementTime = block.timestamp;
```
jackchuma: nit: `block.timestamp` is fixed for the transaction, so disbursementTime is only an alias used to assign storage and emit the event. Assign `lastDisbursementTime = block.timestamp` and emit `block.timestamp` directly to remove the temporary variable

src/L2/FeeDisburser.sol
jackchuma: nit: `feeBalance` is only used for the immediately following zero-balance check. Nothing can change this contract’s balance between the assignment and condition, so `if (address(this).balance == 0)` removes a local without changing behavior

src/L2/FeeDisburser.sol
0xth4nh: Fixed.

src/L2/FeeDisburser.sol
0xth4nh: Fixed.

jackchuma (approved):

PR #393, opened 2026-08-05
Merge main into l3-changes and temporarily pin scheduleId

dguenther:
Merges main into l3-changes and temporarily pin scheduleId so that the hash strategy matches between the base/base prover and the contracts. I think some future fixes are coming for this in base/base, so this is just to get it working for now.


dguenther:
Closing, reviewed out-of-band by @robriks and cherry-picked onto l3-changes.

PR #395, opened 2026-08-09
feat(L1): deploy hinted nitro validator stack

leopoldjoy:
## Summary
- add a standalone Solidity 0.8.26 deployment script for `P384Verifier -> CertManager -> NitroValidator`
- use named direct `lib/...` imports and scoped 10,000-run compiler restrictions matching upstream
- require explicit nonzero CertManager owner and revoker arguments
- test deployment wiring, root-cache initialization, role assignment, and EIP-170 runtime sizes

## Behavior
The script is opt-in and does not change the standard system deployment. `TEEProverRegistry` continues to use the configured legacy `NitroEnclaveVerifier` until C3.

## Testing
- clean `just test --match-path test/deploy/DeployNitroValidatorStack.t.sol` (3 passed)
- targeted script size build: `CertManager` 23,784 bytes with 792 bytes of EIP-170 margin
- CI-equivalent lint, build, snapshots, and full test workflow (1,268 passed, 1 skipped)
- `just snapshots` with no snapshot diff

## Stack
- depends on #394
- base branch is `feat/nitro-validator-c1`, not `main`

PR #394, opened 2026-08-09
chore(L1): pin hinted nitro validator

leopoldjoy:
## Summary
- pin `base/nitro-validator` at `0ea0d12366b4fa44f9e07e4755f2ad36561cb674`
- keep audit and deployment/build integration in the dependency and follow-up deployment PRs

## Testing
- `just deps`
- `just lint-check`
- `just forge-build`

This PR intentionally contains only the dependency pin. The deployment and EIP-170 validation are handled in the stacked C2 PR.

PR #399, opened 2026-08-11
feat(L1): validate hinted Nitro attestations

leopoldjoy:
## Summary
- replace the Registry legacy NitroEnclaveVerifier immutable and two-argument ZK registration API with INitroValidator and registerSigner(bytes,bytes,bytes)
- enforce Base freshness, 48-byte non-debug PCR0, and 65-byte uncompressed secp256k1 public-key policy after hinted validation
- preserve proxy storage, signer/proposer behavior, TEEVerifier behavior, and the legacy verifier deployment for rollback
- validate predeployed Registry immutables before ProxyAdmin upgrades

## Deployment
- add nitroValidator to deploy config, SystemDeploy inputs/outputs, artifacts, and development scripts
- SystemDeploy.upgrade accepts a predeployed TEEProverRegistry implementation, validates its validator and factory immutables, and upgrades the existing proxy
- the legacy NitroEnclaveVerifier proofSubmitter is deliberately left unchanged; existing deployments remain rollback-ready and fresh deployments must prepare it before a legacy rollback
- checked-in nitroValidator config values remain zero until the C2 stack is deployed for each environment; populate the address before a multiproof deployment or upgrade

## Testing
- real AWS-signed hinted attestation through P384Verifier, CertManager, NitroValidator, and the Registry wrapper
- mocked dependency-revert propagation plus real-validator integration and gas coverage
- owner/manager authorization, freshness boundaries, PCR0/debug policy, key format and signer derivation
- standard ProxyAdmin upgrade from a 0.5.0 Registry with owner, manager, game type, proposers, signer set, and image hash preserved
- invalid embedded-validator rejection for prebuilt Registry upgrades
- full suite: 1,286 passed, 0 failed, 1 skipped
- just snapshots and just semver-lock
- source size check: TEEProverRegistry runtime 8,298 bytes

## Stack
- #394 and #395 are merged
- targets main

### DIFF
--- a/interfaces/L1/proofs/tee/ITEEProverRegistry.sol
+++ b/interfaces/L1/proofs/tee/ITEEProverRegistry.sol
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+import { NitroValidator } from "lib/nitro-validator/src/NitroValidator.sol";
+
 import { IDisputeGameFactory } from "interfaces/L1/proofs/IDisputeGameFactory.sol";
-import { INitroEnclaveVerifier } from "interfaces/L1/proofs/tee/INitroEnclaveVerifier.sol";
 import { GameType } from "src/libraries/bridge/Types.sol";
 
 interface ITEEProverRegistry {
-    function NITRO_VERIFIER() external view returns (INitroEnclaveVerifier);
+    function NITRO_VALIDATOR() external view returns (NitroValidator);
     function DISPUTE_GAME_FACTORY() external view returns (IDisputeGameFactory);
     function gameType() external view returns (GameType);
     function isRegisteredSigner(address signer) external view returns (bool);
@@ -17,7 +18,7 @@ interface ITEEProverRegistry {
 
     function setProposer(address proposer, bool isValid) external;
     function setGameType(GameType gameType_) external;
-    function registerSigner(bytes calldata output, bytes calldata proofBytes) external;
+    function registerSigner(bytes calldata attestationTbs, bytes calldata signature, bytes calldata hints) external;
     function deregisterSigner(address signer) external;
     function isValidSigner(address signer) external view returns (bool);
     function getRegisteredSigners() external view returns (address[] memory);
--- a/interfaces/L1/proofs/tee/ITEEProverRegistry.sol
+++ b/interfaces/L1/proofs/tee/ITEEProverRegistry.sol
@@ -1,13 +1,12 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import { NitroValidator } from "lib/nitro-validator/src/NitroValidator.sol";
-
 import { IDisputeGameFactory } from "interfaces/L1/proofs/IDisputeGameFactory.sol";
+import { INitroValidator } from "interfaces/L1/proofs/tee/INitroValidator.sol";
 import { GameType } from "src/libraries/bridge/Types.sol";
 
 interface ITEEProverRegistry {
-    function NITRO_VALIDATOR() external view returns (NitroValidator);
+    function NITRO_VALIDATOR() external view returns (INitroValidator);
     function DISPUTE_GAME_FACTORY() external view returns (IDisputeGameFactory);
     function gameType() external view returns (GameType);
     function isRegisteredSigner(address signer) external view returns (bool);
--- a/scripts/deploy/SystemDeploy.s.sol
+++ b/scripts/deploy/SystemDeploy.s.sol
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.15;
 
-import { VmSafe } from "lib/forge-std/src/Vm.sol";
 import { console2 as console } from "lib/forge-std/src/console2.sol";
+import { NitroValidator } from "lib/nitro-validator/src/NitroValidator.sol";
 import { Script } from "lib/forge-std/src/Script.sol";
+import { VmSafe } from "lib/forge-std/src/Vm.sol";
 
 import { Artifacts } from "scripts/Artifacts.s.sol";
 import { Config } from "scripts/libraries/Config.sol";
@@ -27,6 +28,7 @@ import { IDelayedWETH } from "interfaces/L1/proofs/IDelayedWETH.sol";
 import { IDisputeGame } from "interfaces/L1/proofs/IDisputeGame.sol";
 import { IDisputeGameFactory } from "interfaces/L1/proofs/IDisputeGameFactory.sol";
 import { IVerifier } from "interfaces/L1/proofs/IVerifier.sol";
+import { INitroEnclaveVerifier } from "interfaces/L1/proofs/tee/INitroEnclaveVerifier.sol";
 import { ITEEProverRegistry } from "interfaces/L1/proofs/tee/ITEEProverRegistry.sol";
 import { IOptimismMintableERC20Factory } from "interfaces/universal/IOptimismMintableERC20Factory.sol";
 import { IProxy } from "interfaces/universal/IProxy.sol";
@@ -36,7 +38,6 @@ import { AddressManager } from "src/legacy/AddressManager.sol";
 import { AggregateVerifier } from "src/L1/proofs/AggregateVerifier.sol";
 import { TEEProverRegistry } from "src/L1/proofs/tee/TEEProverRegistry.sol";
 import { TEEVerifier } from "src/L1/proofs/tee/TEEVerifier.sol";
-import { INitroEnclaveVerifier } from "interfaces/L1/proofs/tee/INitroEnclaveVerifier.sol";
 import { ISP1Verifier } from "interfaces/L1/proofs/zk/ISP1Verifier.sol";
 import { ZKVerifier } from "src/L1/proofs/zk/ZKVerifier.sol";
 import { Constants } from "src/libraries/Constants.sol";
@@ -78,6 +79,7 @@ contract SystemDeploy is Script {
         bytes32 multiproofConfigHash;
         uint256 multiproofGameType;
         address nitroEnclaveVerifier;
+        address nitroValidator;
         AggregateVerifier.ScheduleConfig scheduleConfig;
         uint256 multiproofBlockInterval;
         uint256 multiproofIntermediateBlockInterval;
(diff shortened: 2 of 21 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

interfaces/L1/proofs/tee/ITEEProverRegistry.sol
```diff
@@ -1,12 +1,13 @@
...
 
+import { NitroValidator } from "lib/nitro-validator/src/NitroValidator.sol";
+
 import { IDisputeGameFactory } from "interfaces/L1/proofs/IDisputeGameFactory.sol";
-import { INitroEnclaveVerifier } from "interfaces/L1/proofs/tee/INitroEnclaveVerifier.sol";
 import { GameType } from "src/libraries/bridge/Types.sol";
 
 interface ITEEProverRegistry {
-    function NITRO_VERIFIER() external view returns (INitroEnclaveVerifier);
+    function NITRO_VALIDATOR() external view returns (NitroValidator);
```
jackchuma: Can we replace this return value with an interface?

scripts/deploy/SystemDeploy.s.sol
```diff
@@ -202,6 +204,29 @@ contract SystemDeploy is Script {
         _saveUpgradeArtifacts(output_);
     }
 
+    /// @notice Deploys and saves a TEEProverRegistry implementation for an existing OP Chain.
+    /// @param _systemConfigProxy The existing chain's SystemConfig proxy.
+    /// @return output_ The deployed TEEProverRegistry implementation.
+    function deployTEEProverRegistryImplementation(ISystemConfig _systemConfigProxy)
```
jackchuma: Why is this function here and why is it needed? I only see it used in testing

scripts/deploy/SystemDeploy.s.sol
```diff
@@ -709,6 +736,13 @@ contract SystemDeploy is Script {
         if (address(currentGameImpl) == address(0)) return;
 
         if (_impls.teeProverRegistryImpl != address(0)) {
+            TEEProverRegistry newTEEProverRegistry = TEEProverRegistry(_impls.teeProverRegistryImpl);
+            DeployUtils.assertValidContractAddress(address(newTEEProverRegistry));
```
jackchuma: This is redudant. `_upgradeTo` already asserts valid contract address below

scripts/deploy/SystemDeploy.s.sol
```diff
@@ -709,6 +736,13 @@ contract SystemDeploy is Script {
...
+            DeployUtils.assertValidContractAddress(address(newTEEProverRegistry.NITRO_VALIDATOR()));
```
jackchuma: This is also redundant. The nitro validator input is already checked elsewhere

scripts/deploy/SystemDeploy.s.sol
jackchuma: Actually please move this to the `_assertValidImplementations` function and dedup

scripts/deploy/SystemDeploy.s.sol
jackchuma: This could also probably live in `_assertValidImplementations`

scripts/libraries/Types.sol
```diff
@@ -62,6 +64,7 @@ library Types {
         IVerifier zkVerifier;
         INitroEnclaveVerifier nitroEnclaveVerifier;
         ISP1Verifier sp1Verifier;
+        NitroValidator nitroValidator;
```
jackchuma: Every field on this struct is an interface except for what you added. Please use an interface

scripts/multiproof/DeployRiscZeroStack.s.sol
jackchuma, line 1: Might as well just remove this, no?

src/L1/proofs/tee/TEEProverRegistry.sol
```diff
@@ -95,9 +99,9 @@ contract TEEProverRegistry is OwnableManagedUpgradeable, ISemver {
     /// @notice Thrown when setting a game type whose AggregateVerifier has no TEE_IMAGE_HASH.
     error InvalidGameType();
 
-    constructor(INitroEnclaveVerifier nitroVerifier, IDisputeGameFactory factory) {
+    constructor(NitroValidator nitroValidator, IDisputeGameFactory factory) {
```
jackchuma: Interface please

src/L1/proofs/tee/TEEProverRegistry.sol
```diff
@@ -130,36 +134,46 @@ contract TEEProverRegistry is OwnableManagedUpgradeable, ISemver {
...
+    /// @param hints Offchain inverse hints for the attestation signature.
+    function registerSigner(
+        bytes calldata attestationTbs,
+        bytes calldata signature,
+        bytes calldata hints
+    )
+        external
+        onlyOwnerOrManager
+    {
+        bytes memory tbs = attestationTbs;
```
jackchuma, line 159: Do we need to copy this into memory? Why is this line here?

test/L1/proofs/TEEProverRegistry.t.sol
```diff
@@ -395,4 +410,442 @@ contract TEEProverRegistryTest is Test {
...
+    }
+
+    function test_registerSigner_expiredLeafCertificate_reverts() public {
+        _setUpRealAttestation();
+        Fixture memory fixture = _prepareFixture();
+        vm.warp(uint256(fixture.leafNotAfter) + 1);
+
+        vm.expectRevert("cert expired");
+        teeProverRegistry.registerSigner(fixture.tbs, fixture.signature, fixture.hints);
+    }
```
jackchuma: These tests exercise hint parsing, certificate caching, revocation, and expiration failures inside `NitroValidator`/`CertManager`; the registry only propagates those reverts. The mocked propagation test at L538-547 already covers the registry boundary, and the real-attestation test at L656-669 covers integration with the deployed validator stack. Drop the six pass-through failure cases and leave those lower-level branches to the dependency’s test suite

test/L1/proofs/TEEProverRegistry.t.sol
```diff
@@ -395,4 +410,442 @@ contract TEEProverRegistryTest is Test {
...
+    function setUp() public {
+        owner = makeAddr("owner");
+        manager = makeAddr("manager");
+        signer = vm.addr(SIGNER_PRIVATE_KEY);
+
+        bytes memory pcr0 = new bytes(48);
+        for (uint256 i = 0; i < pcr0.length; i++) {
+            pcr0[i] = bytes1(uint8(i + 1));
+        }
+        pcr0Hash = keccak256(pcr0);
```
jackchuma: The loop constructs the same 48 bytes every run, which obscures a constant fixture. Replace it with a single literal:

```solidity
  bytes memory pcr0 = hex"0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f30";
```

  This keeps the test data explicit and removes the setup loop

test/L1/proofs/TEEProverRegistry.t.sol
leopoldjoy: Dropped the six dependency pass-through cases while retaining mocked revert propagation and the real-validator integration/gas test.

interfaces/L1/proofs/tee/ITEEProverRegistry.sol
leopoldjoy: Added a local `INitroValidator` and switched the Registry getter to return it.

scripts/multiproof/DeployRiscZeroStack.s.sol
leopoldjoy: Removed this in the stacked C4 cleanup PR #400.

src/L1/proofs/tee/TEEProverRegistry.sol
leopoldjoy: Switched the Registry immutable and constructor parameter to `INitroValidator`.

test/L1/proofs/TEEProverRegistry.t.sol
leopoldjoy: Replaced the setup loop with the explicit 48-byte PCR0 literal.

src/L1/proofs/tee/TEEProverRegistry.sol
leopoldjoy: Kept the copy because the returned CBOR offsets are consumed by memory-only helpers, and added a comment explaining it.

scripts/libraries/Types.sol
leopoldjoy: Switched `Types.DeployOutput.nitroValidator` to the new `INitroValidator` interface.

scripts/deploy/SystemDeploy.s.sol
leopoldjoy: Moved the implementation code check into `_assertValidImplementations` so it is centralised before any getters are called.

scripts/deploy/SystemDeploy.s.sol
leopoldjoy: Moved this into `_assertValidImplementations`; it remains necessary for prebuilt upgrade implementations, which bypass the fresh-deploy input check.

scripts/deploy/SystemDeploy.s.sol
leopoldjoy: This is the operator entrypoint for deploying a chain-specific implementation against an existing `SystemConfig`; I kept it and documented the Forge invocation.

jackchuma: Why do we need to deploy a chain-specific implementation against an existing SystemConfig? When would we do that from this repo?

leopoldjoy: You are right; removed it and kept predeployed-implementation upgrade coverage.

mw2000 (approved):

PR #401, opened 2026-08-12
feat(L1): make deposit resource minimum base fee deploy-configurable

dguenther:
## Summary
- Add optional `resourceConfigMinimumBaseFee` to deploy config (default `1 gwei`) and wire it into `SystemConfig` resource config initialization.
- Initialize `ResourceMetering.prevBaseFee` from the configured minimum instead of hardcoding `1 gwei`, keeping portal metering aligned with the parent base-fee conversion floor.
- Regenerate `snapshots/semver-lock.json` for the updated OptimismPortal2 init bytecode.

## Test plan
- [x] `just test --match-path test/deploy/ResourceConfigMinimumBaseFee.t.sol`
- [x] `just test --match-path test/deploy/DeployConfig.t.sol`
- [x] `just test --match-path test/L1/ResourceMetering.t.sol`
- [x] `just test --match-path test/deploy/SystemDeploy.t.sol`
- [x] `just semver-lock`

## L3 deploy config
Set `"resourceConfigMinimumBaseFee": 10000000` to align the `0.01 gwei` resource minimum with the portal conversion floor.

### DIFF
--- a/test/deploy/ResourceConfigMinimumBaseFee.t.sol
+++ b/test/deploy/ResourceConfigMinimumBaseFee.t.sol
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { Test } from "lib/forge-std/src/Test.sol";
+
+import { SystemDeploy } from "scripts/deploy/SystemDeploy.s.sol";
+import { Types } from "scripts/libraries/Types.sol";
+import { Constants } from "src/libraries/Constants.sol";
+import { Hash, Proposal } from "src/libraries/bridge/Types.sol";
+
+import { IL1StandardBridge } from "interfaces/L1/IL1StandardBridge.sol";
+import { IOptimismPortal2 } from "interfaces/L1/IOptimismPortal2.sol";
+import { IResourceMetering } from "interfaces/L1/IResourceMetering.sol";
+import { ISystemConfig } from "interfaces/L1/ISystemConfig.sol";
+import { ISP1Verifier } from "interfaces/L1/proofs/zk/ISP1Verifier.sol";
+
+/// @title ResourceConfigMinimumBaseFee_Test
+/// @notice Tests deploy-time configuration of the deposit resource minimum base fee.
+contract ResourceConfigMinimumBaseFee_Test is Test {
+    SystemDeploy internal systemDeploy;
+
+    address internal owner = address(this);
+    address internal guardian = makeAddr("guardian");
+    address internal incidentResponder = makeAddr("incidentResponder");
+    address internal batcher = makeAddr("batcher");
+    address internal unsafeBlockSigner = makeAddr("unsafeBlockSigner");
+    address internal proposer = makeAddr("proposer");
+    address internal challenger = makeAddr("challenger");
+
+    uint256 internal l2ChainId = 901;
+    uint32 internal constant L3_MINIMUM_BASE_FEE = 10_000_000;
+    uint256 internal constant BASE_TX_GAS_CAP = 16_777_216;
+
+    function setUp() public {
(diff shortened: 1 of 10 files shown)


test/deploy/ResourceConfigMinimumBaseFee.t.sol
```diff
@@ -0,0 +1,134 @@
...
+    address internal guardian = makeAddr("guardian");
+    address internal incidentResponder = makeAddr("incidentResponder");
+    address internal batcher = makeAddr("batcher");
+    address internal unsafeBlockSigner = makeAddr("unsafeBlockSigner");
+    address internal proposer = makeAddr("proposer");
+    address internal challenger = makeAddr("challenger");
+
+    uint256 internal l2ChainId = 901;
+    uint32 internal constant L3_MINIMUM_BASE_FEE = 10_000_000;
+    uint256 internal constant BASE_TX_GAS_CAP = 16_777_216;
```
robriks, line 33: could we reference Fusaka here in a doc comment? some chains don't implement the newer tx gas limit, and a future hard fork alters this value

test/deploy/ResourceConfigMinimumBaseFee.t.sol
```diff
@@ -0,0 +1,134 @@
...
+        address depositor = vm.addr(depositorKey);
+        vm.deal(depositor, 10 ether);
+
+        vm.startBroadcast(depositorKey);
+        uint256 gasBefore = gasleft();
+        l1StandardBridge.depositETH{ value: 1 ether }(200_000, hex"");
+        uint256 gasUsed = gasBefore - gasleft();
+        vm.stopBroadcast();
+
+        assertLt(gasUsed, BASE_TX_GAS_CAP);
```
robriks: could we expand this test or add another regression one asserting the issue itself? claude's suggestion:

```solidity
    // 1 gwei default + parent base fee at the 0.01 gwei conversion floor => ~100x burn
    SystemDeploy.DeployOutput memory output = systemDeploy.deploy(_deployInput(uint32(1 gwei)));
    // ... resolve bridge/portal ...
    vm.fee(0.01 gwei);
    // ... measure gasUsed for the same depositETH call ...
    assertGe(gasUsed, BASE_TX_GAS_CAP); // proves the regression the fix prevents
```

robriks (approved):
wow this touches some really involved mechanisms. 

got pretty nerdsniped on how these gas configs are significant and am aligned that this the least invasive way to resolve the 100x overcharge for our use of this codebase as an L3. 

long term, assuming the L3 goes to production, we might need to consider addressing this more invasively if gas conditions (on Base) are volatile **and** we expect to see high deposit volume. 

that's because the 0.01gwei alignment between the minimum base fee and the conversion floor (enacted by this PR) can still be pushed out of balance if Base gas fees drop very low and deposit volume skyrockets.

accordingly, I think the long term solution would be to bound the burn itself so it's provably below the parent's per‑tx cap for the max deposit size. (Claude suggests to clamp gasCost or equivalently cap maximumBaseFee so _amount_max × maximumBaseFee / floor < 2^24). that way deposits would stay includable (ie don't revert) no matter what, with a downstream outcome that the fusaka limit becomes the canonical maximum fee to bridge to L3. 

doing that would be pretty involved though since we'd definitely need to ascertain that bounding the burn doesn't create a cheaper‑spam or undercharge‑exploit surfaces. not worth the effort at this time but wanted to surface it as my recommendation if we get to that point

oh and also, Claude suggested that we bump @custom:semver from 5.2.0 to 5.2.1 for this , I don't think we strictly need to in this PR but whenever we are "done" with the contract changes I think we should think about how to handle the semver so that it's clear our L3 versioning is distinct from canonical Base L2 versioning

robriks (approved):

PR #402, opened 2026-08-14
feat(L3): add attested withdrawal redemption

hughy:
## Summary
- Add `attestedWithdraw` to `L2ToL1MessagePasser`. It records a chain-bound ETH authorization in MessagePasser storage for enclave inclusion checks.
- Add `redeemAttestedWithdrawal` to `OptimismPortal2`. It validates the domain-separated raw signature from a registered TEE signer, prevents replay, honors the Portal pause state, and pays ETH from Portal escrow.
- Wire the TEE registry for new multiproof deployments and Portal upgrades. Update interfaces and generated ABI, storage-layout, and semver snapshots.

## Test plan
- [x] `just test`
- [x] `just snapshots-check-no-build`
- [x] `just semver-lock`
- [x] `forge fmt --check`
- [x] `git diff --check`

### DIFF
--- a/interfaces/L1/IOptimismPortal2.sol
+++ b/interfaces/L1/IOptimismPortal2.sol
@@ -38,7 +38,7 @@ interface IOptimismPortal2 is IProxyAdminOwnedBase {
     error OptimismPortal_InvalidAttestedWithdrawalSignature();
     error OptimismPortal_InvalidAttestedWithdrawalSigner(address signer);
     error OptimismPortal_TEEProverRegistryAlreadySet();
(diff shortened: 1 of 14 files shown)


dguenther (approved):

robriks (approved):

robriks:
I believe since `authHash` isn't bound to the portal contract and/or chain deployment, we can replay authorizations across chain restarts. that seems fine to me for now and probably wouldn't need any changes even when the chain goes non-ephemeral, just wanted to surface it

PR #412, opened 2026-08-21
fix(checks): restore reinitializer checker

### LINKED ISSUE #410: just check and just pre-pr fail: scripts/checks/reinitializer does not exist
`just check` runs `reinitializer-check-no-build`, which shells out to a package that isn't in the tree:

```make
# justfile:245
reinitializer-check-no-build:
  go run ./scripts/checks/reinitializer
```

`scripts/checks/` contains `common`, `interfaces`, `spacers` and `test-validation` — there is no `reinitializer`. The sibling targets use the same relative form (`go run ./scripts/checks/interfaces`) and those packages do exist, so this one looks left behind rather than pathed differently.

## Why it matters

`check` invokes the sub-checks as one `just` command, so the failure stops the chain:

```make
check:
  @just semgrep-test-validity-check \
  semgrep \
  lint-check \
  snapshots-check-no-build \
  unused-imports-check-no-build \
  validate-deploy-configs \
  validate-spacers-no-build \
  reinitializer-check-no-build \   # <- fails here
  interfaces-check-no-build \      # <- never runs
  lint-forge-tests-check-no-build  # <- never runs
```

So two checks that do exist never get a chance to run.

It also breaks `just pre-pr` (and `just pre-commit`, which aliases it). That recipe is `set -e` and calls `just check` partway through:

```bash
just lint
just build-source
just check          # <- aborts here

# Restore build artifacts after running checks.
if [ -d "$TEMP_BUILD_DIR" ]; then
  ...
```

The artifact-restore block after it is skipped too, so the dev-build cache the recipe went to some trouble to save isn't put back.

## Why it isn't showing up in CI

CI doesn't call `check`, `pre-pr` or `pre-commit` — the workflows use `just deps`, `just forge-build`, `just lint-check`, `just semver-lock` and `just test`. So CI stays green while the local "run all checks" path is broken.

## On the check itself

Both targets describe it as "Checks that all upgrade/initialize functions have proper reinitializer modifiers". Grepping the tree, nothing else implements that: the only `reinitializer` hits are OpenZeppelin's modifier in `src/vendor/Initializable.sol` and a comment in `test/universal/OptimismMintableERC20Factory.t.sol`. So the check the justfile advertises isn't running anywhere right now.

## Fix

I haven't sent a PR because the right fix depends on something I can't tell from outside: whether the script was dropped deliberately (in which case `reinitializer-check`, `reinitializer-check-no-build` and the line in `check` should go) or whether it went missing and should come back. Happy to send whichever you'd prefer.

Verified against `main` at `4f7acda`, and `scripts/checks/` on GitHub shows the same four packages.

Mabolla:
## Summary

- restore the missing `scripts/checks/reinitializer` checker referenced by `just reinitializer-check-no-build`
- adapt the checker to use this repository's local `scripts/checks/common` package
- restore the local `just check` / `just pre-pr` path without changing the existing recipes

## Why

`just check` currently invokes `go run ./scripts/checks/reinitializer`, but that package is absent from the tree, so the check chain aborts before later checks can run. The checker existed when the contracts were imported into this repository and the equivalent implementation still exists upstream in Optimism, which indicates the missing package is the regression rather than the `justfile` target.

Closes #410.

## Validation

- `git diff upstream/main...HEAD --check`
- verified the branch differs from `main` by exactly one added file: `scripts/checks/reinitializer/main.go`
- local `go test ./scripts/checks/reinitializer` could not be run because Go is not installed in the Windows environment used for this contribution

### DIFF
--- a/scripts/checks/reinitializer/main.go
+++ b/scripts/checks/reinitializer/main.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"fmt"
+	"math"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/base/contracts/scripts/checks/common"
+	"github.com/ethereum-optimism/optimism/op-chain-ops/solc"
+)
+
+func main() {
+	if _, err := common.ProcessFilesGlob(
+		[]string{"forge-artifacts/**/*.json"},
+		[]string{},
+		processFile,
+	); err != nil {
+		fmt.Printf("Error: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func processFile(path string) (*common.Void, []error) {
+	artifact, err := common.ReadForgeArtifact(path)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	if err := checkArtifact(artifact); err != nil {
+		return nil, []error{err}
+	}
+
+	return nil, nil
+}
+
+func checkArtifact(artifact *solc.ForgeArtifact) error {
+	// Skip interfaces.
+	if strings.HasPrefix(artifact.Ast.AbsolutePath, "interfaces/") {
+		return nil
+	}
+
+	// Skip if we have no upgrade function.
+	upgradeFn := getFunctionByName(artifact, "upgrade")
+	if upgradeFn == nil {
+		return nil
+	}
+
+	// We can have an upgrade function without an initialize function.
+	initializeFn := getFunctionByName(artifact, "initialize")
+	if initializeFn == nil {
+		return nil
+	}
+
+	// Grab the reinitializer value from the upgrade function.
+	upgradeFnReinitializerValue, err := getReinitializerValue(upgradeFn)
+	if err != nil {
+		return fmt.Errorf("error getting reinitializer value from upgrade function: %w", err)
+	}
+
+	// Grab the reinitializer value from the initialize function.
+	initializeFnReinitializerValue, err := getReinitializerValue(initializeFn)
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #433, opened 2026-09-08
feat(L1): bake the Cobalt speedup activation into AggregateVerifier

0xth4nh:
Follow-up to #431 (merged).

## What changed?

`AggregateVerifier` no longer reads the fast-block activation from `ProtocolVersions`.

- `ScheduleConfig` gains `uint64 fastBlockActivationTimestamp`, exposed as `uint64 public immutable FAST_BLOCK_ACTIVATION_TIMESTAMP`.
- `FAST_BLOCK_UPGRADE_INDEX` (the hardcoded `getSchedule()[12]` lookup) is deleted. `_firstFastBlock()` now derives the boundary from the immutable and the L2 genesis/block-time immutables.
- `0` means "speedup unscheduled" — every game uses the slow-block intervals. This lets the implementation ship before the activation time is agreed, at the cost of a redeploy once it is.
- Deploy path gains a `multiproofFastBlockActivationTimestamp` config key (defaults to `0`), plumbed through `DeployConfig` → `SystemDeploy` with a `uint64` overflow check.
- Semver `0.2.0` → `0.3.0`; `snapshots/semver-lock.json` regenerated (only the `AggregateVerifier` entry moves).

`PROTOCOL_VERSIONS` is deliberately retained — `activatedScheduleId(claimTimestamp)` still drives scheduleId pinning. Only the interval selection stops depending on the registry.

## Why?

`ProtocolVersions` will not be activated for Cobalt. It runs in metrics-only mode on sepolia and mainnet (`--upgrade-signal.mode`), so index 12 of `getSchedule()` is never written on those chains.

The old lookup **fails open**: an index past the end of the schedule returns `type(uint256).max`, so `_firstFastBlock()` sits beyond every real block number and the verifier keeps handing out slow-block intervals forever. On a chain that has already switched to 200ms blocks, that is a silent mismatch between the proposer's cadence and the verifier's expectation — no revert, no event, just wrong intervals.

An immutable also removes a class of race the registry has: every game backed by a given implementation resolves the same boundary, and there is no mutation that can move an in-flight game across it. Changing the activation means deploying a new implementation, which is a new game type in the factory, so it can never apply retroactively.

## How to test?

```
just build-go-ffi
forge test --match-contract AggregateVerifier
```

The new case is the one that pins the behaviour change:

```
forge test --match-test test_intervalsForStartingBlock_activationIndependentOfRegistry_succeeds -vv
```

It writes a `ProtocolVersions` schedule that stops one entry *short* of index 12 — i.e. a chain that never signalled the speedup — asserts `getSchedule().length == 12`, then deploys a verifier with `fastBlockActivationTimestamp = L2_GENESIS_TIMESTAMP + 100` and checks the boundary lands at block 50 (`divUp(100, 2)`): block 49 gets the slow interval pair, block 50 gets the fast pair. On `main` this test cannot pass — with the registry short of index 12 the verifier stays slow at every block.

Full suite on this branch: `1232 passed, 0 failed, 1 skipped`.

Deploy-time check: `multiproofFastBlockActivationTimestamp` is absent from every existing config, so `readUintOr(..., 0)` yields `0` and existing deployments keep slow intervals — `forge test --match-contract SystemDeploy` covers that path unchanged.

### DIFF
--- a/interfaces/L1/proofs/IAggregateVerifier.sol
+++ b/interfaces/L1/proofs/IAggregateVerifier.sol
@@ -29,6 +29,7 @@ interface IAggregateVerifier is IDisputeGame {
     function L2_GENESIS_BLOCK_NUMBER() external view returns (uint256);
     function L2_GENESIS_TIMESTAMP() external view returns (uint64);
     function L2_BLOCK_TIME() external view returns (uint64);
+    function FAST_BLOCK_ACTIVATION_TIMESTAMP() external view returns (uint64);
     function SLOW_BLOCK_INTERVAL() external view returns (uint256);
     function SLOW_INTERMEDIATE_BLOCK_INTERVAL() external view returns (uint256);
     function FAST_BLOCK_INTERVAL() external view returns (uint256);
--- a/scripts/deploy/DeployConfig.s.sol
+++ b/scripts/deploy/DeployConfig.s.sol
@@ -52,6 +52,7 @@ contract DeployConfig is Script {
     uint256 public l2OutputOracleStartingBlockNumber;
     uint256 public l2OutputOracleStartingTimestamp;
     uint256 public multiproofSlowBlockInterval;
+    uint256 public multiproofFastBlockActivationTimestamp;
     uint256 public multiproofFastBlockInterval;
     uint256 public multiproofFastIntermediateBlockInterval;
     uint256 public multiproofGameType;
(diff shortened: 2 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0xth4nh:
Closing — we're keeping ProtocolVersions as the single source for the activation time rather than introducing a second one.

